PR #2, opened 2020-11-12
Fix gotobuffer

JimboKearn:
Fixed GotoBuffer to go to a terminal which
is already closed And made it vim compatible.

### DIFF
--- a/plugin/harpoon.vim
+++ b/plugin/harpoon.vim
@@ -21,6 +21,14 @@ fun! GotoBuffer(ctrlId)
     end
 
     let bufh = l:contents[1]
+    if !bufexists(bufh)
+        " Create the terminal
+        terminal
+        call SetBuffer(a:ctrlId)
+    endif
+
+    let contents = g:win_ctrl_buf_list[a:ctrlId]
+    let bufh = l:contents[1]
     call nvim_win_set_buf(0, l:bufh)
 endfun
 
@@ -55,7 +63,3 @@ fun! SendTerminalCommand(ctrlId, command)
     let job_id = l:contents[0]
     call chansend(l:job_id, a:command)
 endfun
-
-
-
-
--- a/plugin/harpoon.vim
+++ b/plugin/harpoon.vim
@@ -12,7 +12,7 @@ fun! GotoBuffer(ctrlId)
     let contents = g:win_ctrl_buf_list[a:ctrlId]
     if type(l:contents) != v:t_list
         " Create the terminal
-        terminal
+        exe has("nvim") ? "terminal" : "terminal ++curwin"
         call SetBuffer(a:ctrlId)
     end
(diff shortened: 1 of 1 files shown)


plugin/harpoon.vim
```diff
@@ -34,7 +42,7 @@ fun! SetBuffer(ctrlId)
         return
     end
 
-    let g:win_ctrl_buf_list[a:ctrlId] = [b:terminal_job_id, nvim_win_get_buf(0)]
```
ThePrimeagen, line 37: So vim operates on simply the buffer handle and thats it?

Interesting.

plugin/harpoon.vim
```diff
@@ -21,11 +21,19 @@ fun! GotoBuffer(ctrlId)
...
+    endif
+
+    let contents = g:win_ctrl_buf_list[a:ctrlId]
+    let bufh = l:contents[1]
+    exe "b" . l:bufh
 endfun
 
 fun! SetBuffer(ctrlId)
-    if has_key(b:, "terminal_job_id") == 0
+    if &buftype != "terminal"
```
ThePrimeagen, line 36: What is `buftype` and what does `&` do?

Sorry I am very terrible at vimL.

plugin/harpoon.vim
```diff
@@ -12,7 +12,7 @@ fun! GotoBuffer(ctrlId)
     let contents = g:win_ctrl_buf_list[a:ctrlId]
     if type(l:contents) != v:t_list
         " Create the terminal
-        terminal
+        exe has("nvim") ? "terminal" : "terminal ++curwin"
```
ThePrimeagen, line 15: what is `++curwin`?

ThePrimeagen (approved):

ThePrimeagen:
I totally approve your PR, just more confused as to what some of these VimL
things mean!  Could you enlighten me and I'll approve aftewords?

@JimboKearn

plugin/harpoon.vim
JimboKearn: Normally if you do a `:terminal` in vim it will create the terminal
in a split. So you can give it the `++curwin` flag to create the
terminal in the current window. So it works likes the neovim terminal.

plugin/harpoon.vim
JimboKearn: I didn't get the question.

ThePrimeagen: just looking at the code.  I notice that you use `bufnr()` to grab an id and to communicate with the terminal.  So vim terminal just uses buffer id to do communication?

JimboKearn:
This is my first open source contribution! I learned vim through your videos, Thx :)

plugin/harpoon.vim
JimboKearn: Yeah, it only need terminal buffer id to do communication
Vim doesn't use any kind of terminal id for communication

PR #3, opened 2021-02-19
expand ~ for terminal locations

polarmutex:


ThePrimeagen:
Gave you the cred.  Moved that sweet sweet code you just wrote into the harpoon#init.lua setup call.  That way it will also happen for mark.lua file

PR #5, opened 2021-02-22
fix: nil

### LINKED ISSUE #4: Lua error after installing harpoon
After installing `harpoon`, I get the following error when running `:bnext`, `:bprev`, `:CocCommand explorer` etc...
```
Error detected while processing BufLeave Autocommands for "*":
E5108: Error executing lua ...onfig/nvim/autoload/plugged/harpoon/lua/harpoon/mark.lua:34: attempt to get length of field 'marks' (a nil value)
```
I'm using Neovim nightly `v0.5.0-dev+1097-g18f90336c`.

ThePrimeagen:
This _may_ fix #4 

Its my only guess is that there is a poorly hydrated config, therefore, I will enforce it is correct

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -30,6 +30,24 @@ local M = {}
 --]]
 harpoon_config = harpoon_config or {}
 
+function ensure_correct_config(config) 
+    local projects = config.projects
+    if projects[cwd] == nil then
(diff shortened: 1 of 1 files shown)


ThePrimeagen:
This may work for @gpcureton  's issue.

PR #6, opened 2021-02-22
More nil work

ThePrimeagen:
#4 should definitely be solved at this point


ThePrimeagen:
@polarmutex is there something I am missing.  tbl_deep_extend does not operate the way you would think

Issue #11, opened 2021-02-22
add config option to make file/terminal open in a split window and/or floating window


polarmutex:

ThePrimeagen:
I like where this is going!

ThePrimeagen:
So I was thinking about doing this.  @polarmutex are you wanting to take a crack at it?  

I think this is an excellent opportunity to make the "ui" file really handle all things UI related.  ui.lua should really just use terminal / marks as a state manager and do the ui work.  

Still trying figure out which direction the arrows should point.  (marks < - ui makes sense.  term < - > ui (not sure yet)). 

Any who, lmk.

P.S.  The more Vim specific functions we can take out of terminal / marks the better as they will become very straight forward state machines that can be tested.

ThePrimeagen:
I was also going to start doing this next if you don't.  So don't feelz bad :)

polarmutex:
@ThePrimeagen I have not started looking at it yet. I can if you want to start looking at other changes

ziontee113:
Hi, I just created an autocmd for opening file in vsplit:
```lua
local group = vim.api.nvim_create_augroup("Harpoon Augroup", { clear = true })
vim.api.nvim_create_autocmd("FileType", {
    pattern = "harpoon",
    group = group,
    callback = function()
        vim.keymap.set("n", "<C-V>", function()
            local curline = vim.api.nvim_get_current_line()
            local working_directory = vim.fn.getcwd() .. "/"
            vim.cmd("vs")
            vim.cmd("e " .. working_directory .. curline)
        end, { noremap = true, silent = true })
    end,
})
```

bryant-the-coder:
> Hi, I just created an autocmd for opening file in vsplit:
> 
> ```lua
> local group = vim.api.nvim_create_augroup("Harpoon Augroup", { clear = true })
> vim.api.nvim_create_autocmd("FileType", {
>     pattern = "harpoon",
>     group = group,
>     callback = function()
>         vim.keymap.set("n", "<C-V>", function()
>             local curline = vim.api.nvim_get_current_line()
>             local working_directory = vim.fn.getcwd() .. "/"
>             vim.cmd("vs")
>             vim.cmd("e " .. working_directory .. curline)
>         end, { noremap = true, silent = true })
>     end,
> })
> ```
> 
> I don't know about compatibility with older version of nvim tho.

interesting. might add this to my fork :)

PR #13, opened 2021-02-23
Lil sumptin

ThePrimeagen:
@polarmutex what do you think about this config for #11 ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,64 +23,6 @@ Plug 'ThePrimeagen/harpoon'
 ```
 
 ## Harpooning
+-- TODO -- Sorry (still being dev'd on)
 
-### The Basics
-Harpoon provides up to 4 slots for terminals.  Now you don't want to overload
-terminals with everything.  I just tend to think about this as project specific
-tasks.  Such as, `yarn lint`, `yarn test`, or some command based off of current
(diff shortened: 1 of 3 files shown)


lua/harpoon/init.lua
```diff
@@ -8,22 +8,32 @@ local cache_config = string.format("%s/harpoon.json", data_path)
 local M = {}
 
 --[[
+-- type Layout = "float" | "split" | "vsplit" | "hsplit" | "replace" (default) | "tab"
```
ThePrimeagen, line 11: @polarmutex I am thinking that this would allow for every option.

polarmutex: do we need spilt with hsplit and vsplit? or what is the difference between them?

polarmutex (commented):
config looks good, just the  one question

README.md
```diff
@@ -23,64 +23,6 @@ Plug 'ThePrimeagen/harpoon'
 ```
 
 ## Harpooning
+-- TODO -- Sorry (still being dev'd on)
 
-### The Basics
```
brandoncc, line 27: It seems like this readme is a bad idea to merge 😉

Issue #10, opened 2021-02-22
Create Replace At function for Marks


ThePrimeagen:
Great suggestion by Asbjorn to have a `replace_at(idx: number): void` 

This should be a really simple function to write and a great first issue if anyone wants to take this.

jesseleite:
What should happen if you try to `replace_at(3)`, if only 1 mark exists in the table?

PR #14, opened 2021-02-23
Add set_current_at(idx) function

asbjornhaland:
I have some struggles with showing the marks in the window some times. I have not quite figured out when or what is causing this. After discussion with ThePrimeagen, I'm making this PR now so we can have a look at it.

#10

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -85,6 +85,7 @@ function expand_dir(config)
 end
 
 M.save = function()
+    require('harpoon.mark').fill_missing_indexes()
     Path:new(cache_config):write(vim.fn.json_encode(harpoon_config), 'w')
 end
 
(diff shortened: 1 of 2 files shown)


asbjornhaland:
If marks is an empty table and you set current buf to `4`, you have this mark config:
```
{
  marks = {                                 
    [4] = "include/articletemplates/std_template.php"
  }
}
```
This _is_ a valid lua table, however, when we try to save (with `vim.fn.json_encode()`), viml doesn't know how to convert the missing indexes and errors: `5100: Cannot convert given lua table: table should either have a sequence of positive integer keys or contain only string keys.`.

Current solution is to call `fill_missing_indexes()` in `save()` to fill the missing indexes with `vim.NIL` (`nil` is never inserted into a table).

asbjornhaland:
Some functions now use `#config.marks` (for instance `toggle_quick_menu()`.
The given table:
```
{
  marks = {                                 
    [4] = "include/articletemplates/std_template.php"
  }
}
```
returns 0. We should think about how we loop the config since there can be 'holes'

asbjornhaland:
This is dead.

PR #15, opened 2021-02-23
feat(save): remove dead code to prevent nil value on save

asbjornhaland:
`get_config()` is removed and replaced by
`get_term_config()` and `get_mark_config()`

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -85,16 +85,6 @@ function expand_dir(config)
 end
 
 M.save = function()
-    local term_config = terminals.get_config()
-    local mark_config = mark.get_config()
-
-    if not harpoon_config.projects[cwd] then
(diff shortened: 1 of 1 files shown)


lua/harpoon/init.lua
```diff
@@ -43,15 +43,15 @@ function merge_table_impl(t1, t2)
     end
 end
 
-function merge_tables(...)
```
ThePrimeagen, line 46: Is my remove whitespace not being called?  Double you Tee Eff!!! let me fix that.

ThePrimeagen:
and rubber stamp

ThePrimeagen:
And actually rubber stamp

lua/harpoon/init.lua
asbjornhaland: Seems like it. I did remove in several files.

PR #18, opened 2021-03-04
feat: floating windows and notifications!

anott03:


lua/harpoon/ui.lua
```diff
@@ -63,6 +63,51 @@ M.nav_file = function(id)
...
+    vim.api.nvim_buf_set_lines(info.bufnr, 0, 5, false, {"!!! Notification", text})
+    vim.api.nvim_set_current_win(prev_win)
 
+    return {
+        bufnr = info.bufnr,
+        win_id = info.win_id
+    }
+end
+
+function M.close_notification(bufnr)
```
ThePrimeagen, line 108: I like where this is going.  But looking at notifications.  It seems like we should probably create this into a class such that notifications can be closed / timedout / whatever in a self contained way

For the sake of this pr.  I like this :7

PR #20, opened 2021-03-04
undefined global mark_config

rudotriton:
I believe it was meant to swap the values in the same config

Someone reported this in discord

```
E5108: Error executing lua /home/peter/.vim/plugged/harpoon/lua/harpoon/mark.lua:51: attempt to index global 'mark_config' (a nil value)
```


ThePrimeagen:
Cheers!

ThePrimeagen:
Huge thanks @rudotriton !! Thanks so much for this

PR #26, opened 2021-03-15
Set filetype for quick menu

brandoncc:
This adds the harpoon filetype to the quick menu so that mappings can be setup specifically for use in the menu. Runonsentence+300points.


ThePrimeagen:
this is great

PR #27, opened 2021-03-15
Save changes when closing menu

brandoncc:
This PR adds auto-saving of changes when the quick menu is closed. The implementation is very naive. It accepts any changes, and grabs the text after the first space in each line. No validations included.

I also added a .editorconfig and the dev reloader you helped me with.


ThePrimeagen:
I'll merge, just answer questions

PR #28, opened 2021-03-15
Add telescope finder

brandoncc:
@ThePrimeagen what do you think about baking in a telescope finder?

### DIFF
--- a/lua/harpoon/telescope.lua
+++ b/lua/harpoon/telescope.lua
@@ -1,0 +1,43 @@
+-- telescope.lua
+--
+-- A customizable finder for telescope.nvim
+-- (https://github.com/nvim-telescope/telescope.nvim)
+--
+-- Usage: :lua require("harpoon.telescope").marks({results_title = 'My Title'})
+local previewers = require('telescope.previewers')
+local pickers = require('telescope.pickers')
+local finders = require('telescope.finders')
+local sorters = require('telescope.sorters')
+local harpoon = require('harpoon')
+
+local default_options = {
+    results_title = 'Harpoon',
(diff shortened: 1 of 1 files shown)


lua/harpoon/telescope.lua
```diff
@@ -0,0 +1,43 @@
...
+-- Usage: :lua require("harpoon.telescope").marks({results_title = 'My Title'})
+local previewers = require('telescope.previewers')
+local pickers = require('telescope.pickers')
+local finders = require('telescope.finders')
+local sorters = require('telescope.sorters')
+local harpoon = require('harpoon')
+
+local default_options = {
+    results_title = 'Harpoon',
+    finder = finders.new_table(harpoon.get_mark_config().marks),
```
ThePrimeagen, line 15: Probably want to do this in the function.  Reason being that marks is not being updated

ThePrimeagen (changes requested):
do that and you win

lua/harpoon/telescope.lua
brandoncc: That is a really good point. I didn't think about this when refactoring to allow overriding the options. I don't think it makes sense to allow overriding the finder anyway, now that I think of it.

Also, you said "you win". Were you avoiding adding a telescope integration since you have the quick menu? If that is the case, I have no problem just closing this.

brandoncc: The issue should be resolved, but I'm totally fine if we don't merge this. Please don't merge it unless it is something you want in the project @ThePrimeagen

brandoncc:
Closing as it should be moved to a separate package.

PR #31, opened 2021-03-15
Dont mark nil or empty string

brandoncc:


lua/harpoon/mark.lua
```diff
@@ -74,14 +74,19 @@ local function swap(a_idx, b_idx)
...
+M.add_file = function(file_name_or_buf_id)
+    local buf_name = get_buf_name(file_name_or_buf_id)
 
     if valid_index(get_index_of(buf_name)) then
         -- we don't alter file layout.
         return
     end
 
+    if buf_name == "" or buf_name == nil then
+        print("Couldn't find a valid file name to mark, sorry.")
```
ThePrimeagen, line 86: Lets drop this, if we use add_file programmatically it could get a bit goofy.

brandoncc: Okay, but it seems like even if you use it programmatically, you would _never_ want to insert an empty string or a nil... From what I can tell, empty string or nil don't work right when trying to access them. Am I missing something?

PR #33, opened 2021-03-16
Add file toggling

brandoncc:
@ThePrimeagen this PR adds file toggling and also updates rm_file to allow passing in the same arguments that add_file accepts.


ThePrimeagen:
@brandoncc this looks good.  We should extract the file is marked check from `add_file` into a function and just use that instead of `add_file` to call to see if it adds.

brandoncc:
I was thinking the same thing, but didn't do it because it just works 😂 I'll make that change.

PR #34, opened 2021-03-16
feat(mark)!: add set_current_at

### LINKED ISSUE #10: Create Replace At function for Marks

asbjornhaland:
BREAKING CHANGE:
- filter_nulls(list) renamed to filter_empty_string(list)
- swap(a, b) removed
- trim() removed
- promote() removed
- promote_to_front() removed
- remove_nils() removed
- shorten_list() removed

Fix #10

### DIFF
--- a/lua/harpoon/mark.lua
+++ b/lua/harpoon/mark.lua
@@ -1,45 +1,12 @@
-local Path = require('plenary.path')
 local harpoon = require('harpoon')
 local utils = require('harpoon.utils')
 
 local M = {}
 
-local function valid_index(idx)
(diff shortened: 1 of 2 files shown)


asbjornhaland:
Doesn't seem like I can add reviewers, so I'll ping you here.

Like we discussed the promote, swap and shorten methods are removed and set_current_at is added.

@ThePrimeagen @brandoncc

PR #35, opened 2021-03-17
Improve checking if marks exist

brandoncc:


lua/harpoon/mark.lua
```diff
@@ -31,6 +31,25 @@ local function get_buf_name(id)
     return ""
 end
 
+local function mark_exists(buf_name)
+    local marks = harpoon.get_mark_config().marks
+
+    for idx = 1, #marks do
```
asbjornhaland, line 37: I believe we have to use `M.get_length()` instead of `#marks` because of the index problem we discussed with prime.

brandoncc: Oh, I didn't realize that. Thanks @asbjornhaland !

brandoncc: Should be resolved now :-)

lua/harpoon/mark.lua
```diff
@@ -31,6 +31,25 @@ local function get_buf_name(id)
...
+            return true
+        end
+    end
+
+    return false
+end
+
+local function validate_buf_name(buf_name)
+    if buf_name == "" or buf_name == nil then
+        error("Couldn't find a valid file name to mark, sorry.")
```
ThePrimeagen, line 48: We do need to come up with a logging strategy.  We don't want to over due it to the end user, but I do like it during alpha!  

We can revisit this later.

lua/harpoon/mark.lua
```diff
@@ -178,16 +194,15 @@ M.set_mark_list = function(new_list)
 end
 
 M.toggle_file = function(file_name_or_buf_id)
-    local mark_count_before = #harpoon.get_mark_config().marks
-
-    M.add_file(file_name_or_buf_id)
+    local buf_name = get_buf_name(file_name_or_buf_id)
 
-    local mark_count_after = #harpoon.get_mark_config().marks
+    validate_buf_name(buf_name)
```
ThePrimeagen, line 199: What's the point of this?  It has no early return, it will simply just print an error message.

lua/harpoon/mark.lua
```diff
@@ -73,10 +92,7 @@ M.add_file = function(file_name_or_buf_id)
         return
     end
 
-    if buf_name == "" or buf_name == nil then
-        error("Couldn't find a valid file name to mark, sorry.")
-        return
-    end
+    validate_buf_name(buf_name)
```
ThePrimeagen, line 95: You dropped the early return

ThePrimeagen: Also above, it wont let me comment, but 5 lines above there is 

```
    if M.valid_index(M.get_index_of(buf_name)) then
        -- we don't alter file layout.
        return
    end
```

That should be replaced with this logic

```
if not validate_buf_name(buf_name) or mark_exists(buf_name) then
    return
end
```

ThePrimeagen (changes requested):

lua/harpoon/mark.lua
brandoncc: I believe `error` raises an error and the code stops execution (unless the error is handled). If we were using `print`, the execution would continue.

lua/harpoon/mark.lua
brandoncc: I didn't want to have to repeat this checking logic all over, that was the reason I made the validation function which stops execution if validation fails. As for the early return, it error raises so execution stops on its own.

PR #38, opened 2021-03-17
Use autocommand for saving buffer changes

brandoncc:


brandoncc:
@ThePrimeagen this should close #38. There is one thing I experienced that we might need to deal with. If you navigate to a file within the harpoon buffer on accident, then close, you save the file lines as your mark content.

This bug is already in master, so not new here.

Also, I found a bug. `win_info.bufh` is nil, so we should be using `win_info.bufnr`. I fixed that as well.

ThePrimeagen:
The more I think about it I become torn between this and `:w` being the way you settle your list.

brandoncc:
I agree that `:w` is better and would love to get that into this PR. I left some comments on the issue you opened. If you have any ideas how to get past the issues, I'd love to hear them.

PR #39, opened 2021-03-18
feat(ui): add BufWriteCmd to handle ":w" in quick menu

### LINKED ISSUE #36: Closing quick menu doesn't save unless toggle is called
Autocmd time!  Lets listen for win close events and ensure that we save upon floating window closed with ZZ/q

asbjornhaland:
Fix #36

@ThePrimeagen @brandoncc Can you have a look, please test this before merge.

### DIFF
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -68,13 +68,12 @@ M.toggle_quick_menu = function()
     vim.api.nvim_buf_set_option(bufh, "filetype", "harpoon")
     vim.api.nvim_buf_set_option(bufh, "buftype", "acwrite")
     vim.api.nvim_buf_set_option(bufh, "bufhidden", "delete")
-    vim.api.nvim_buf_set_option(bufh, "modified", false)
     vim.cmd(string.format("autocmd BufWriteCmd <buffer=%s> :lua require('harpoon.ui').on_menu_save()", bufh))
+    vim.cmd(string.format("autocmd BufModifiedSet <buffer=%s> set nomodified", bufh))
 end
 
 M.on_menu_save = function()
     save_changes()
-    vim.cmd("set nomodified")
 end
 
 M.nav_file = function(id)
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -1,4 +1,3 @@
-local Path = require('plenary.path')
 local float = require('plenary.window.float')
 local Marked = require('harpoon.mark')
 
@@ -68,6 +67,7 @@ M.toggle_quick_menu = function()
     vim.api.nvim_buf_set_lines(bufh, 0, #contents, false, contents)
     vim.api.nvim_buf_set_option(bufh, "filetype", "harpoon")
     vim.api.nvim_buf_set_option(bufh, "buftype", "acwrite")
+    vim.api.nvim_buf_set_option(bufh, "bufhidden", "delete")
     vim.cmd(string.format("autocmd BufWriteCmd <buffer=%s> :lua require('harpoon.ui').on_menu_save()", bufh))
 end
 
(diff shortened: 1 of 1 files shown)


lua/harpoon/ui.lua
```diff
@@ -65,8 +64,16 @@ M.toggle_quick_menu = function()
...
+    vim.api.nvim_buf_set_name(bufh, "harpoon-menu")
     vim.api.nvim_buf_set_lines(bufh, 0, #contents, false, contents)
     vim.api.nvim_buf_set_option(bufh, "filetype", "harpoon")
+    vim.api.nvim_buf_set_option(bufh, "buftype", "acwrite")
+    vim.cmd(string.format("autocmd BufWriteCmd <buffer=%s> :lua require('harpoon.ui').on_menu_save()", bufh))
+end
+
+M.on_menu_save = function()
+    save_changes()
+    vim.cmd("set nomodified")
```
brandoncc, line 76: Interesting, so this is the piece I was missing?

lua/harpoon/ui.lua
```diff
@@ -65,8 +64,16 @@ M.toggle_quick_menu = function()
         contents[idx] = string.format("%d %s", idx, file)
     end
 
+    vim.api.nvim_buf_set_name(bufh, "harpoon-menu")
```
brandoncc, line 67: Oh, I think this is really the one I was missing!

lua/harpoon/ui.lua
asbjornhaland: If modified is not reset, you indicate the write was *not* successful, + you get error when closing nvim that you have modified buffers.

You were very close though, I think your missing piece was naming the buffer with `vim.api.nvim_buf_set_name`.

lua/harpoon/ui.lua
asbjornhaland: Yes

ThePrimeagen:
@brandoncc and @asbjornhaland have you verified this all works?  I'll merge it (no time to test currently)

Also, very interesting.  Could you write up exactly / point me to which :h commands to read on, so I understand why this works :)

asbjornhaland:
- `:h E676` is the explanation for `buftype=acwrite` and `BufWriteCmd`
- `:h E32` is nameing the buffer

brandoncc:
I get this when I try to reopen the menu after doing `ZZ`:

```
E5108: Error executing lua /Users/brandoncc/.vim/plugged/harpoon/lua/harpoon/ui.lua:67: Vim:E95: Buffer with this name already exists
```

Maybe we need to recycle the buffer when opening by trying to find a buffer by the same name instead of creating it? Only create if there isn't one already by the same name?

brandoncc:
@asbjornhaland I found some issues.

`:wq` and `ZZ` work. `:w` and then `:q` doesn't, though.

Also, if you make a change and then `:q`, it tells you that you need to save your changes. I wonder if we need to set modified to false in an `on_lines` callback?

asbjornhaland:
If you make a change, doesn't it make sense that you need to `:q!` - this is normal vim  behavior?

I will have a look at why modified is set after a `:w` - it shouldn't.

brandoncc:
> If you make a change, doesn't it make sense that you need to `:q!` - this is normal vim behavior?
> 
> I will have a look at why modified is set after a `:w` - it shouldn't.

I might have misunderstood @ThePrimeagen, but I thought he said he didn't want to require `:q!` to close. At first I thought he was saying he _did_ want it, but then I thought he meant he didn't.

Sorry, I wasn't clear about my bug report. `:wq` saves, but `:w` and then `:q` doesn't. The old content comes back after doing `:w`,`:q`.

ThePrimeagen:
I _do not want_ q! required

brandoncc:
Scratch that `:w:q` bug report, it looks like it was user error.

PR #41, opened 2021-03-21
Added nav_next and nav_prev functions

Bhanukamax:
This allows to cycle through the harpoon marks list
Hi @ThePrimeagen , not sure if this is something you want.

Also I haven't done much lua+neovim plugin stuff, so please give me any feedback, if I'm doing it totally wrong, let me know too :slightly_smiling_face: 

sample key binding:
```
nnoremap <silent> <C-Left> :lua require("harpoon.ui").nav_prev()<CR>
nnoremap <silent> <C-Down> :lua require("harpoon.ui").nav_next()<CR>
```

### DIFF
--- a/lua/harpoon/mark.lua
+++ b/lua/harpoon/mark.lua
@@ -203,5 +203,9 @@ end
 
 M.to_quickfix_list()
 
+M.get_current_index = function()
+    return M.get_index_of(vim.fn.bufname(vim.fn.bufnr()))
+end
+
 return M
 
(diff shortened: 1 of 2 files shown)


ThePrimeagen:
!

I was going to make this a feature, but you did it :)

One thing.  I do want to state this.  Harpoon is meant to be thought -> action -> location.  this does fundamentally "break" that notion since you go back into search mode.

nrupatunga:
Is there a option to cycle through the list?

PR #42, opened 2021-03-23
Offsets

ThePrimeagen:
THE OFFSETS


ThePrimeagen:
@brandoncc @asbjornhaland if there is anything I have missed. lmk.

So easy to miss things :)

brandoncc:
Nothing jumps out at me. I'll try it out and see if I can find any bugs. Nice work!

plugin/mark.vim
```diff
@@ -1,5 +1,10 @@
+fun! HarpoonLeave()
```
ThePrimeagen, line 1: At some point, we could probably kill this file and the other .vim file.

Just use vim.cmd [[ ... ]]

lua/harpoon/mark.lua
```diff
@@ -103,41 +112,51 @@ M.add_file = function(file_name_or_buf_id)
...
-    end
+    local ok, res = pcall(function()
+        local buf_name = get_buf_name()
+        local idx = M.get_index_of(buf_name)
+        if not M.valid_index(idx) then
+            return
+        end
 
-    local line = vim.api.nvim_eval("line('.')");
+        harpoon.get_mark_config().marks[idx].row =
```
asbjornhaland, line 152: Are `col` ever set? If it's not needed, I think it should be removed from `create_mark()`?

asbjornhaland (commented):
I'm not available to test this right now, but could not see anything that stood out (except my one comment) when I skimmed through the code. Good work!

plugin/mark.vim
asbjornhaland: I agree.

lua/harpoon/mark.lua
ThePrimeagen: I want to get col to work, I just haven't yet.  I am going to keep it in there until I figure that part out :)

asbjornhaland: `:lua print(vim.fn.col("."))`

PR #48, opened 2021-03-24
Fix clearing project config

jesseleite:
Thank you for all the 🍐 's. They taste wonderful.


ThePrimeagen:
Hey @polarmutex oops :)

I should of thought about that `projects[k] = nil` thing a bit more...  My bad :)

polarmutex:
sorry, definitely did not think through that one. Lesson learned: do not make pull requests while watching prime

jesseleite:
> sorry

Are you Canadian too @polarmutex?  No sorries unless you are Canadian, this is alpha.

Issue #49, opened 2021-03-27
toggle_quick_menu does not save the config


A-Daneel:
issue description: quick menu ordering does not update when using `toggle_quick_menu`. Using `:wq` does update it.

nvim version: 0.5.0-dev+1157-g0ab88c2ea
harpoon version: up to date as of 2021/03/27

`get_mark_config` before changing the order:
```
{                                                                                                            
  marks = { {
      col = 0,
      filename = "tests/file1.js",
      row = 43
    }, {
      col = 0,
      filename = "tests/file2.js",
      row = 19
    }, {
      col = 0,
      filename = "tests/data.json",
      row = 81
    }, {
      col = 0,
      filename = "tests/data.json_backup",
      row = 23
    } }
}
```

after changing the 2nd and 3th item around and close using `toggle_quick_menu` nothing has changed.

asbjornhaland:
@ThePrimeagen: Can be closed after #50, right?

ThePrimeagen:
I am going to close it because #50 should of addressed it.

PR #50, opened 2021-03-28
fix(ui): menusave now gets called on menu toggle

A-Daneel:
Not sure if this is the proper way to solve this.
But my moving the son_menu_save call into the closing if-loop it will
get properly called on when closing the menu with toggle_quick_menu().

I think merging this change will resolve issue #49.

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -46,7 +46,7 @@ end
 
 local function merge_tables(...)
     local out = {}
-    for i = 2, select("#",...) do
+    for i = 1, select("#",...) do
         merge_table_impl(out, select(i, ...))
     end
     return out
@@ -129,7 +129,7 @@ M.setup = function(config)
 
     local complete_config =
         merge_tables(
-            {projects = {}},
+            {projects = {} , global_settings = {["save_on_toggle"] = false}},
             expand_dir(c_config),
             expand_dir(u_config),
             expand_dir(config))
@@ -139,6 +139,10 @@ M.setup = function(config)
     ensure_correct_config(complete_config)
 
     HarpoonConfig = complete_config
+end
+
+M.get_global_settings = function()
+    return HarpoonConfig.global_settings
 end
 
 M.get_term_config = function()
@@ -155,7 +159,7 @@ M.print_config = function()
 end
 
 -- Sets a default config with no values
-M.setup({projects = {}})
+M.setup()
 
 return M
 
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -41,6 +41,11 @@ end
 
 M.toggle_quick_menu = function()
     if win_id ~= nil and vim.api.nvim_win_is_valid(win_id) then
+        local global_config = harpoon.get_global_settings()
+        print(global_config.save_on_menu_quit)
+        if global_config.save_on_toggle == true then
+            vim.cmd(":lua require('harpoon.ui').on_menu_save()")
+        end
         vim.api.nvim_win_close(win_id, true)
 
         win_id = nil


lua/harpoon/ui.lua
```diff
@@ -80,7 +81,7 @@ M.toggle_quick_menu = function()
     vim.api.nvim_buf_set_option(bufh, "filetype", "harpoon")
     vim.api.nvim_buf_set_option(bufh, "buftype", "acwrite")
     vim.api.nvim_buf_set_option(bufh, "bufhidden", "delete")
-    vim.cmd(string.format("autocmd BufWriteCmd <buffer=%s> :lua require('harpoon.ui').on_menu_save()", bufh))
+    --not sure if this line is still needed now?..
     vim.cmd(string.format("autocmd BufModifiedSet <buffer=%s> set nomodified", bufh))
```
asbjornhaland, line 85: Both these au are used for `:wq` and `:w`.

asbjornhaland (changes requested):

lua/harpoon/ui.lua
A-Daneel: Then would it be logical to move these to lines to just before we call `vim.api.nvim_win_close(win_id, true) ` on line 56?
The problem seems to me that those two lines aren't called on closing.
```
vim.cmd(string.format("autocmd BufWriteCmd <buffer=%s> :lua require('harpoon.ui').on_menu_save()", bufh))
vim.cmd(string.format("autocmd BufModifiedSet <buffer=%s> set nomodified", bufh))
```

brandoncc:
This is how the saving used to work. It was intentionally changed to match the semantics of saving in vim. `:w` and `:wq` should both save for you, while `:q` will close the menu but _not_ save your changes. This is how @ThePrimeagen wanted it to function.

If `w` is not saving for you, please open an issue about that.

A-Daneel:
So what I perceived as an error is it's intended behavior. Should I close my issue and this PR?

ThePrimeagen:
Sorry for being out on this.  Let me review now.

ThePrimeagen:
@A-Daneel I have an idea

Could you add an option to the top level config, defaults to false

Check out [project config](https://github.com/ThePrimeagen/harpoon/blob/master/lua/harpoon/init.lua#L13)
It has a top level key `projects`, it could have other top level keys as well.  Imagine `ui` and you could add a key that is `saveOnWrite`, which defaults to `true`.

Thoughts? @asbjornhaland

asbjornhaland:
If we should have 'autosave' I think the best solution is as @ThePrimeagen mentions - an option in the config.

brandoncc:
I also agree that an option for this is a good idea.

ThePrimeagen:
@A-Daneel I may take a stab at this at some point, are you going to follow up on this?  Or should I consider this a task for me?

A-Daneel:
Kinda depends on how much of a priority it is for you. Think between study and work it'll take me a day to figure out what is going on where. If it's no rush I'd love to take take a stab at it.

A-Daneel:
I've done a very small change. if you manually edit your harpoon.json to match [project config](https://github.com/ThePrimeagen/harpoon/blob/master/lua/harpoon/init.lua#L13) it only saves on close if `save_on_menu_quit` is true.

Feedback is more then welcome

A-Daneel:
Not sure how to proceed from here, and if what I did thus far is correct.

@ThePrimeagen @asbjornhaland could either of you tell me if I'm on the right track with this.

ThePrimeagen:
@A-Daneel I am so sorry.  I didn't see this!

Let me review!

lua/harpoon/init.lua
```diff
@@ -159,7 +167,14 @@ M.print_config = function()
 end
 
 -- Sets a default config with no values
-M.setup({projects = {}})
+M.setup(
```
ThePrimeagen, line 170: Lets change setup to be called with nothing 

```
M.setup()
```

Then lets change the defaults inside the function instead of here since they will get wiped out with any call to setup.

lua/harpoon/init.lua
```diff
@@ -129,6 +132,7 @@ M.setup = function(config)
 
     local complete_config =
         merge_tables(
+            {global_settings = {}},
```
ThePrimeagen, line 135: Lets combine global_settings + projects table into one.

```
local complete_config =
        merge_tables(
            {projects = {}, global_settings = {["save_on_toggle"] = false}},
            ...
```

There is a bug here that will likely bite you.  You can save this by changing the idx from starting at 2 in merge_tables (i have a fix in a different branch).

This will make that option default off, but you can turn it on by calling setup!

lua/harpoon/init.lua
```diff
@@ -10,6 +10,9 @@ local M = {}
 
 --[[
 {
+    global_setting = {
+        ["save_on_menu_quit"] = false
```
ThePrimeagen, line 14: You can remove this.  I am getting ready to create the documents for harpoon

lua/harpoon/init.lua
ThePrimeagen: Also the change to merge tables should be like this

```
local function merge_tables(...)
    local out = {}
    for i = 1, select("#",...) do // IT WAS for i = 2, select("#",...) 
        merge_table_impl(out, select(i, ...))
    end
    return out
end
```

A-Daneel: something here goes wrong I think, it keeps initiating it as true. Feels like I'm missing something obvious, will push the test of the changes and take another look after dinner.

lua/harpoon/ui.lua
```diff
@@ -53,6 +53,11 @@ end
 
 M.toggle_quick_menu = function()
     if win_id ~= nil and vim.api.nvim_win_is_valid(win_id) then
+        local global_config = harpoon.get_global_settings()
+        print(global_config.save_on_menu_quit)
+        if global_config.save_on_toggle == true then
+            vim.cmd(":lua require('harpoon.ui').on_menu_save()")
```
asbjornhaland, line 59: I believe you can just call `save_changes()` here without the `vim.cmd()`.

A-Daneel: that indeed seems to work, not sure why I did it my way in the first place.
Something else, is what I do with writing the global config and then reading the correct property in the if statement the 'correct' way?

ThePrimeagen: hah!  

You can just rewrite this as 
```
if global_config.save_on_toggle then
    require('harpoon.ui').on_menu_save()
end
```

No need to check for truth when its true :)
Also no need to call lua _from lua_ *think about it*

A-Daneel: I already rewrote line 59 to just `save_changes` as Asbjorn suggested. Think it should be all good now. yes?

ThePrimeagen:
@A-Daneel Last thing.

Could you rebase your commits to 1 commit!

git rebase -i HEAD~6   -- I think

A-Daneel:
@ThePrimeagen That was a learning experience and a half. Upside is I now understand rebasing and reverting git history. 

Hope it's ready for merge

PR #51, opened 2021-03-31
Fix bug when adding file on previously removed mark

joacohoyos:


brandoncc:
Thanks for contributing! Can you provide steps to reproduce the issue this is fixing?

joacohoyos:
> Thanks for contributing! Can you provide steps to reproduce the issue this is fixing?

Hey! Sure

- Add 2 file marks
- Delete the first one
- Finally try to add a new one. 

You won't be able to add marks unless you clear them all

lua/harpoon/mark.lua
```diff
@@ -115,7 +115,7 @@ M.add_file = function(file_name_or_buf_id)
     for idx = 1, M.get_length() do
         local filename = M.get_marked_file_name(idx)
         if filename == "" then
-            harpoon.get_mark_config().marks[idx] = create_mark(filename)
```
ThePrimeagen, line 118: Great!

PR #55, opened 2021-04-07
feat(wildcards)

ThePrimeagen:
I haven't tested this yet.  Literally, inspired, created the code real quick

Thoughts @asbjornhaland @brandoncc


brandoncc:
@ThePrimeagen  How would this be used? I don't have any wildcard use experience in lua, so reading the code didn't help me understand the problem it solves or how to use it :-/

PR #57, opened 2021-04-14
remove print statement to a nonexistent value

rudotriton:
I assume it was used for debugging, but the config value was renamed so it prints `nil` every time the menu is opened.


ThePrimeagen:
OHH SNAP!  I thought I removed that.  Several thank yous!

Issue #43, opened 2021-03-23
shutdown -r now 


ThePrimeagen:
I have had the file saving process not work 2x `shutdown -r now`.  Perhaps we should consider on every update to the harpoon mark config.  We could use Path's async save

Thoughts @asbjornhaland @brandoncc

asbjornhaland:
Wow. That's not good. I did not experience any of that throughout my workday today. Does it hang? What did you do when it broke?

ThePrimeagen:
Effectively the previous state the config file was in remains.  The updating doesn't seem to happen, no hanging.

asbjornhaland:
How do you close the window - toggle, `:wq`?

ThePrimeagen:
Ahh!  I see the confusion.

No.  So I have vim running in some terminal.  I have made several changes to the marks.  I then go to another terminal and type `shutdown -r now`.  This will erase any progress made on marks since the vimleave didn't get the time it needed to write a file.

asbjornhaland:
Sounds about right. The object is updated all the time, but is only written to disk on `VimLeavePre`.

asbjornhaland:
What do we think about this now? Should we save to disk more often, or?

PR #58, opened 2021-04-16
Save on change

ThePrimeagen:


ThePrimeagen:
@polarmutex 
@asbjornhaland
@brandoncc 

I keep Losing my file because vim closes before save can complete

asbjornhaland:
Maybe we should save after every interaction. - rm and set_current_at as well.

ThePrimeagen:
@asbjornhaland I love this idea!

Perhaps we should emit an event that way people can hook into it and even create their own actions afterwords

asbjornhaland:
I agree, this sounds good.

asbjornhaland:
BTW this is a duplicate of #43 , right?

ThePrimeagen:
I'll get it all sorted first thing in morning

ThePrimeagen:
there we go.  I think that is everything.  I will be testing it for the day, if all works out, I'll be merging it!

asbjornhaland:
Nice, good work! Should we emit change for `store_offset()` and `toggle_file()` as well?

ThePrimeagen:
@asbjornhaland toggle file does emit.  (it emits from add_file and rm_file)

store_offset... I never really thought about that... that is a good point.

asbjornhaland:
Aah, yes, of course. LGTM, I guess you did not encounter any issues during the work day?

I think this means #43 can be closed.

ThePrimeagen:
yes. it fixed several weirdsnesses I have experienced

PR #59, opened 2021-04-21
fix(mark): move early exit after saving

asbjornhaland:


lua/harpoon/mark.lua
```diff
@@ -9,14 +9,14 @@ local callbacks = {}
...
 local function emit_changed()
-    if not callbacks["changed"] then
-        return
-    end
-
     if harpoon.get_global_settings().save_on_change then
         harpoon.save()
     end
 
+    if not callbacks["changed"] then
```
ThePrimeagen, line 16: I am such an idiot.

PR #65, opened 2021-04-29
Add logging

rudotriton:
Here's my take on logging for #8. It uses the convenient `plenary.log` which writes a `harpoon.log` to `vim.fn.stdpath("cache")`.

By default the logging is enabled for `warning` and above. It can be changed with an env variable, e.g. when launching nvim `$ HARPOON_LOG=trace nvim` or setting `vim.g.harpoon_log_level` before requiring harpoon. I was considering harpoon's `global_setting` instead of the `vim.g`, but you'll get a recursive require if you want to log in `init.lua` then. Maybe there's some clever trick to get around this.

`warn` and `error` were assigned according to the existing print and error statements. Functions that users would directly interact with are `info`s, even just so that finding them in the logs is easier. Most other are `debug`. Local functions are assigned `trace` along with some more verbose stuff such as parts of the config at different stages.

I have already used the logs to debug the harpoon marker for my statusline. While doing that I discovered that the callbacks for the "changed" event weren’t actually called, since the callbacks table is nested, so it was iterating the keys not the callbacks under the keys. I've fixed that here, but if necessary I can submit it as a separate PR.

Over time it needs to be adjusted. I have used it only very briefly now, so I can't really tell what feels out of place. With use we know where we wish it was more verbose and what just creates noise.

---

The readme has been updated now to document logging and I've reverted the nonfunctional changes made by my formatter, although we should consider having a formatter.

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -28,7 +28,6 @@ local M = {}
     ... high level settings
 }
 --]]
-
 HarpoonConfig = HarpoonConfig or {}
 
 -- tbl_deep_extend does not work the way you would think
@@ -48,7 +47,7 @@ end
 
 local function merge_tables(...)
     local out = {}
-    for i = 1, select("#", ...) do
+    for i = 1, select("#",...) do
         merge_table_impl(out, select(i, ...))
     end
     log.trace("_merge_tables(): Output", out)
@@ -62,10 +61,10 @@ local function ensure_correct_config(config)
         log.trace("ensure_correct_config(): No config found for:", vim.loop.cwd())
         projects[vim.loop.cwd()] = {
             mark = {
-                marks = {},
+                marks = {}
             },
             term = {
-                cmds = {},
+                cmds = {}
             },
         }
     end
@@ -86,7 +85,7 @@ local function ensure_correct_config(config)
         local mark = marks[idx]
         if type(mark) == "string" then
             mark = {
-                filename = mark,
+                filename = mark
             }
             marks[idx] = mark
         end
@@ -145,13 +144,17 @@ M.setup = function(config)
         c_config = {}
     end
 
-    local complete_config = merge_tables({
-        projects = {},
-        global_settings = {
-            ["save_on_toggle"] = false,
-            ["save_on_change"] = true,
+    local complete_config =
+        merge_tables({
+            projects = {},
+            global_settings = {
+                ["save_on_toggle"] = false,
+                ["save_on_change"] = true,
+            },
         },
-    }, expand_dir(c_config), expand_dir(u_config), expand_dir(config))
+        expand_dir(c_config),
+        expand_dir(u_config),
+        expand_dir(config))
 
     -- There was this issue where the vim.loop.cwd() didn't have marks or term, but had
     -- an object for vim.loop.cwd()
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -145,16 +145,14 @@ M.setup = function(config)
     end
 
     local complete_config =
-        merge_tables({
-            projects = {},
-            global_settings = {
+        merge_tables(
+            {projects = {} , global_settings = {
                 ["save_on_toggle"] = false,
                 ["save_on_change"] = true,
-            },
-        },
-        expand_dir(c_config),
-        expand_dir(u_config),
-        expand_dir(config))
+            }},
+            expand_dir(c_config),
+            expand_dir(u_config),
+            expand_dir(config))
 
     -- There was this issue where the vim.loop.cwd() didn't have marks or term, but had
     -- an object for vim.loop.cwd()
@@ -193,3 +191,4 @@ end
 M.setup()
 
 return M
+
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -127,12 +127,13 @@ M.setup = function(config)
     end
 
     local ok, u_config = pcall(read_config, user_config)
-    local ok2, c_config = pcall(read_config, cache_config)
 
     if not ok then
         log.debug("setup(): No user config present at", user_config)
         u_config = {}
     end
+
+    local ok2, c_config = pcall(read_config, cache_config)
 
     if not ok2 then
         log.debug("setup(): No cache config present at", cache_config)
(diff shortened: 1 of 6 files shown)


lua/harpoon/mark.lua
```diff
@@ -9,20 +10,24 @@ local callbacks = {}
...
     if harpoon.get_global_settings().save_on_change then
         harpoon.save()
     end
 
     if not callbacks["changed"] then
+        log.trace("_emit_changed(): no callbacks for 'changed', returning")
         return
     end
 
-    for _, cb in pairs(callbacks) do
```
rudotriton, line 20: Is this meant to run callbacks for every event or just for `changed`?

ThePrimeagen:
On the road. I'll review this weekend

ThePrimeagen:
All right I just got back, I'll try to review this first thing on stream tomorrow night.

lua/harpoon/ui.lua
```diff
@@ -47,16 +51,13 @@ local function get_menu_items()
     return indices
 end
 
-local save_changes = function()
-    Marked.set_mark_list(get_menu_items())
-end
-
```
rudotriton, line 53: `toggle_quick_menu` calls `on_menu_change` which would directly call this `save_changes`. Due to commit d0f4ad, there is now no need for this abstraction and `Marked.set_mark_list(get_menu_items())` has been moved to `on_menu_change`.

asbjornhaland:
This looks really promising.

lua/harpoon/init.lua
```diff
@@ -66,11 +71,13 @@ local function ensure_correct_config(config)
 
     local proj = projects[vim.loop.cwd()]
     if proj.mark == nil then
-        proj.mark = {marks = {}}
+        log.trace("ensure_correct_config(): No marks found for", vim.loop.cwd())
+        proj.mark = { marks = {} }
```
ThePrimeagen, line 75: Formatting change.

Is this a formatter you are using or manual?  If there is a formatter you are using, what is the settings?  If there are settings, make the formatting changes + the formatter settings as a PR!!  That way we can just have uniform spaced code.

lua/harpoon/init.lua
```diff
@@ -66,11 +71,13 @@ local function ensure_correct_config(config)
...
     end
 
     if proj.term == nil then
-        proj.term = {cmds = {}}
+        log.trace("ensure_correct_config(): No terminal commands found for", vim.loop.cwd())
+        proj.term = { cmds = {} }
```
ThePrimeagen, line 80: same

lua/harpoon/mark.lua
```diff
@@ -71,25 +84,30 @@ local function create_mark(filename)
...
+        log.error("_validate_buf_name(): Not a valid name for a mark,", buf_name)
         return
     end
 end
 
 M.get_index_of = function(item)
+    log.debug("get_index_of():", item)
     if item == nil then
-        error("You have provided a nil value to Harpoon, please provide a string rep of the file or the file idx.")
+        log.error("get_index_of(): You have provided a nil value to Harpoon, please provide a string rep of the file or the file idx.")
```
ThePrimeagen, line 110: Does this also call `error` itself?  If it doesn't lua execution wont stop.

lua/harpoon/mark.lua
```diff
@@ -135,6 +157,7 @@ end
 
 M.add_file = function(file_name_or_buf_id)
     local buf_name = get_buf_name(file_name_or_buf_id)
+    log.info("add_file():", buf_name)
```
ThePrimeagen, line 160: I am having a hard time following when `info` vs `debug` vs `trace` is used.  Could we have `trace` for all function calls / arguments provided?

lua/harpoon/mark.lua
```diff
@@ -149,8 +172,9 @@ M.add_file = function(file_name_or_buf_id)
     emit_changed();
 end
 
--- dont_emit_on_changed should only be used internally
```
ThePrimeagen, line 152: I hated that!  Thank you for changing it.

lua/harpoon/mark.lua
```diff
@@ -149,8 +172,9 @@ M.add_file = function(file_name_or_buf_id)
...
+-- _emit_on_changed == false should only be used internally
 M.remove_empty_tail = function(_emit_on_changed)
+    log.debug("remove_empty_tail()")
     _emit_on_changed = _emit_on_changed == nil or _emit_on_changed
     local config = harpoon.get_mark_config()
     local found = false
```
ThePrimeagen, line 180: this is literally terrible code.  I should definitely fix this asap ;)

lua/harpoon/term.lua
```diff
@@ -1,24 +1,26 @@
...
     if term_id == nil then
-        -- TODO: Throw an erro?
+        log.error("_create_terminal(): term_id is nil")
+        -- TODO: Throw an error?
         return nil
     end
 
     -- Make sure the term buffer has "hidden" set so it doesn't get thrown
     -- away and cause an error
-    vim.api.nvim_buf_set_option(bufh, 'bufhidden', 'hide')
```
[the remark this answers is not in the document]
ThePrimeagen: 69420 iq

lua/harpoon/ui.lua
```diff
@@ -106,7 +112,13 @@ M.nav_file = function(id)
 
     vim.api.nvim_set_current_buf(buf_id)
     if set_row and mark.row then
-        vim.cmd(string.format(":call cursor(%d, %d)", mark.row, mark.col))
+        local ok, err = pcall(vim.cmd, string.format(":call cursor(%d, %d)", mark.row, mark.col))
```
ThePrimeagen, line 115: can this error?

ThePrimeagen:
@rudotriton I would say the number 1 feedback is consistency.  You had `info`, `debug`, and `trace`.

I think that function calls should be `trace` with their arguments.
I think that debug should be anything that is needed in between function calls.

Do error logs cause `error` calls in lua?  If not, we need to call `error`.

lua/harpoon/ui.lua
rudotriton: technically it can if string.format doesn't receive numbers, but I'm not actually sure if we would run into this. I'll change it around anyway for now.

rudotriton:
I have updated the pr

- added errors next to error logging
- changed all function calls to log.trace and anything in between to log.debug

ThePrimeagen:
Awesome!  This is great!  Thanks for this!

PR #66, opened 2021-05-04
formatting with stylua

rudotriton:
- used [stylua](https://github.com/JohnnyMorganz/StyLua) for consistent formatting
- all changes are non-functional

eventually have it as part of the CI process?


ThePrimeagen:
This looks great.  I don't like the defaults, but we can change that easily.

PR #70, opened 2021-05-10
feat(ui): navigate to mark on enter

asbjornhaland:
It has it's limitations and can be improved.
Should we split the line `lua print(vim.fn.getline('.'))` on space and get the index?
Should we split the line and navigate with the path?

The path is safer if you haven't saved before hitting enter. Thoughts @ThePrimeagen ?

I quickly made this to better my workflow, but can adjust after settling the questions above.

The reasoning behind this is that I might have more marks than 4 marks, and I don't want to delete any.
Workspaces may solve the need for this, but I like that I can navigate with `enter` from normal mode.


ThePrimeagen:
Yes!  I was going to do the same thing.

This is great times!

I feel like this + a couple other improvements and we have ourself the "finished" product.

rudotriton:
Does it also close harpoon's popup for you? For me it opens the mark right there. Should there be a `win_close` or something as well?

Screen Shot

rudotriton:
I open the popup menu, press enter on a mark and it happens. Am I misunderstanding how it is meant to work?

I submitted #76, just in case, as toggling the menu before navigating does fix it for me.

PR #81, opened 2021-05-12
prevent indexing nil vals in ensure_correct_config

rudotriton:
setting a mark non-consecutively may result in `ensure_correct_config` trying to access nil values. `set_current_at` does fix the empty slots eventually, but before it does so, it calls `get_length` which calls `ensure_correct_config`.

It happens only sometimes due to how lua does memory allocation for arrays. Getting the length with `#` may or may not return the last index, depending on where the previous values are located.

---

Other:

some changes to log orders

### DIFF
--- a/lua/harpoon/mark.lua
+++ b/lua/harpoon/mark.lua
@@ -264,10 +264,8 @@ M.get_marked_file_name = function(idx)
 end
 
 M.get_length = function()
-    log.trace("M.get_length()")
-    local length = table.maxn(harpoon.get_mark_config().marks)
-    log.debug("M.get_length():", length)
-    return length
+    log.trace("get_length()")
+    return table.maxn(harpoon.get_mark_config().marks)
 end
 
 M.set_current_at = function(idx)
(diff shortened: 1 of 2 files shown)


lua/harpoon/mark.lua
```diff
@@ -264,16 +264,17 @@ M.get_marked_file_name = function(idx)
 end
 
 M.get_length = function()
-    log.trace("get_length()")
-    return table.maxn(harpoon.get_mark_config().marks)
+    log.trace("M.get_length()")
+    local length = table.maxn(harpoon.get_mark_config().marks)
+    log.debug("M.get_length():", length)
```
ThePrimeagen, line 269: Could we drop the get_len log and just have the one that has the length?

ThePrimeagen (changes requested):

ThePrimeagen:
Other than that, perfect!

rudotriton:
I reverted the change to get_length's logs for now, I believe that maintaining the log order is more important than having that length there.

ThePrimeagen:
We really need to add unit tests for marks. Within the next couple weeks I'll try to take care of that.

PR #84, opened 2021-05-21
Add requirements missing in installation instructions

DanMargera:
Took me a while to figure out that it wouldn't work on nvim 0.4. Also, popup.nvim is required.


ThePrimeagen:
How about making it a subheader under installation? Just make it super obvious. Something like the following.

### installation
#### must use neovim 0.5x

PR #86, opened 2021-05-22
custom terminal creation command

jonathf:
Adding the possibility for terminal handle to be something else than just pure shell.

For example `.gotoTerminal(2, ':e term://ipython')` and `.sendCommand(2, 'print(123)', ':e term://ipython')` ensures that the terminal one jumps to and send commands to, is an Ipython REPL.

### DIFF
--- a/lua/harpoon/term.lua
+++ b/lua/harpoon/term.lua
@@ -4,11 +4,14 @@ local log = require("harpoon.dev").log
 local M = {}
 local terminals = {}
 
-local function create_terminal()
-    log.trace("_create_terminal()")
+local function create_terminal(create_cmd)
+    if not create_cmd then
+        create_cmd = ":terminal"
+    end
+    log.trace("_create_terminal(): Init:", create_cmd)
     local current_id = vim.fn.bufnr()
 
-    vim.cmd(":terminal")
+    vim.cmd(create_cmd)
     local buf_id = vim.fn.bufnr()
     local term_id = vim.b.terminal_job_id
 
@@ -27,11 +30,11 @@ local function create_terminal()
     return buf_id, term_id
 end
 
-local function find_terminal(idx)
+local function find_terminal(idx, create_cmd)
     log.trace("_find_terminal(): Terminal:", idx)
     local term_handle = terminals[idx]
     if not term_handle or not vim.api.nvim_buf_is_valid(term_handle.buf_id) then
-        local buf_id, term_id = create_terminal()
+        local buf_id, term_id = create_terminal(create_cmd)
         if buf_id == nil then
             return
         end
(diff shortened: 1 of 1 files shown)


lua/harpoon/term.lua
```diff
@@ -45,24 +48,27 @@ local function find_terminal(idx)
...
+M.gotoTerminal = function(idx, create_cmd)
     log.trace("gotoTerminal(): Terminal:", idx)
-    local term_handle = find_terminal(idx)
+    local term_handle = find_terminal(idx, create_cmd)
 
     vim.api.nvim_set_current_buf(term_handle.buf_id)
 end
 
-M.sendCommand = function(idx, cmd, ...)
+M.sendCommand = function(idx, cmd, create_cmd)
```
jonathf, line 58: This one had me stumped a bit. The most intuitive would be something like:
```lua
.sendCommand(2, ":e term://ipython", "print('hello world!')")
```
however, this breaks current usage and forces the user to address terminal creation for every command for all users.
So this implementation I am proposing only breaks the variable interpolation part, which still can be used by using a table:
```lua
.sendCommand(2, {"echo %s", "hello world"})
```
I like not breaking user space, but I am unsure how to do that in this case.

jonathf: BTW, I'd like to noe that the interpolation feature makes vim filename expansion and ipython magic commands harder to use. Personally, I'd remove it and let the user do it manually with:
```lua
.sendCommand(2, string.format("echo %s", "hello world"))
```

ThePrimeagen:
I'll be thinking about this. 

My first gut instinct is that we should probably have a different command than send command.  That way we can keep the sprintf feel of sendCommand.

Are there any other use cases other than REPLs?

jonathf:
We can make a new command.
Though there is also another approach to add the shell inits as a configuration string. which defaults to `:terminal`.

Yes, REPLs are not the only use case here. Any terminal application work. Personally, I often grab `htop` to check on my local processes. (With a small pid-tree function, I could even filter it to only include the nvim instance sub-threads.)

ThePrimeagen:
Ok, I ahve an idea.

@jonathf what if we do this.

`sendCommand(number OR table, cmd, ...args).`

if table, it must have `idx`.  If it has the terminal creation type, then it can inject that. 

Thoughts???

ThePrimeagen:
This could also allow us to have something like `killOnExit` added as a flag

jonathf:
That is a good idea. It also opens up for more arguments later. I'll write up the code.

What kind of feature does killOnExit give? Are you thinking of to just not setting the bufhidden hide option?

My though is that without harpoon I'd have:
```vim
:autocmd TermClose * :q
```
to skip the `[Process exited 0]` message prompt.
We could perhaps bake something similar for all terminals with `killOnExit=true`?

jonathf:
@ThePrimeagen, code is updated with support for idx as number and table.

ThePrimeagen:
@jonathf So sorry for forgetting.  Baby + limited time on computers = very forgetful

I'll get this merged asap :)

ThePrimeagen:
@jonathf could you update the formatting?

jonathf:
Fixed now.

ThePrimeagen:
Sorry!  Just checked this.  Been traveling.

PR #87, opened 2021-05-24
Allow statusline marks to differ between windows

rudotriton:
When using split windows, the mark status of the buffer in the active window was shown on every statusline. This change allows inactive windows to query the status with a buffer number to show a relevant index.

---

Also, `vim.fn.bufname()` uses current buffer by default, so it's identical to `vim.fn.bufname(vim.fn.bufnr())`


ThePrimeagen:
great! sorry about the wait :)

PR #92, opened 2021-08-04
feat(term): add clear all functionality for terms

### LINKED ISSUE #91: Feat: Add the ability to clear all terminals
### Situation
 I'm using harpoon alongside with [git worktree](https://github.com/ThePrimeagen/git-worktree.nvim). 

When switching trees i always have to close the terminals I'm using to reopen them in the current directory.

### Feat
 It would be nice to have a function that let you kill all terminal, similar to the clear all marks feature.
 
 I don't mind trying to do it myself, but i don't know if it would be a good feature or not, maybe my case is very specific

trewjames:
Added a function to simply clear/de-harpoon any previously harpooned terminals.
The terminals will still remain in the buffer pool but calling gotoTerminal will open up a fresh terminal.

Closes #91

First time contributing to anything but hope this is fine.

### DIFF
--- a/lua/harpoon/term.lua
+++ b/lua/harpoon/term.lua
@@ -72,4 +72,9 @@ M.sendCommand = function(idx, cmd, ...)
     end
 end
 
+M.clear_all = function()
+    log.trace("clear_all(): Clearing all terminals.")
+    terminals = {}
+end
+
 return M
--- a/lua/harpoon/term.lua
+++ b/lua/harpoon/term.lua
@@ -74,6 +74,9 @@ end
 
 M.clear_all = function()
     log.trace("clear_all(): Clearing all terminals.")
+    for _, term in ipairs(terminals) do
+        vim.api.nvim_buf_delete(term.buf_id, { force = true })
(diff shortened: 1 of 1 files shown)


lua/harpoon/term.lua
```diff
@@ -72,4 +72,9 @@ M.sendCommand = function(idx, cmd, ...)
     end
 end
 
+M.clear_all = function()
+    log.trace("clear_all(): Clearing all terminals.")
```
ThePrimeagen, line 76: I think we should go through each terminal and call `vim.api.nvim_buf_delete`.  That should at least remove it from the buffers menu too!

ThePrimeagen (changes requested):

lua/harpoon/term.lua
trewjames: Got it. My initial thought was to mimick the behavior of `clear_all()` for marks and leave the buffer pool untouched but I'll make that change.

ThePrimeagen:
Thumbs

PR #95, opened 2021-08-19
feat(term): configuration to run command automatically on send

EthanJWright:
When sending a command to the terminal, I want it to immediately start running. This is a configuration implementation to allow for this use case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ require("harpoon").setup({
     global_settings = {
         save_on_toggle = false,
         save_on_change = true,
+        enter_on_sendcmd = false,
     },
     ... your other configs ...
 })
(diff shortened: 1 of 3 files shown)


asbjornhaland:
You can add `\n` in your config to get it to run immediately FYI.

EthanJWright:
You mean in the command I'm passing? I get that, I expected this to be default behavior (you see it in other neovim term plugins)￼ so figured it was worth adding a config for. Open to closing if this is a more personal use case.

pranavrao145:
I actually ran into this same issue without realizing that there was a PR for this. I think this would be really helpful, because this would mean you don't have to go into the harpoon term again, enter insert mode, and hit enter, just to run the command. The way I see it, it doesn't hurt to have this option in the config.

ThePrimeagen:
DID I FORGET TO REVIEW THIS?
 Sorry.  Life crae :)

ThePrimeagen:
BOOM.e  Thanks @EthanJWright

Issue #96, opened 2021-08-29
Dynamic terminal commands... 


RishabhRD:
# The Problem
Preconfigured commands for project helps a lot when we have to test the project or build the project all the time. However, vimrc doesn't seem to be the best location for saving it. Not only it mixes project configuration with vimrc(we can have many temporary projects... long vimrc is unreasonable) also it makes it static. That means with evolution of project, if the workflow changes... user have to change its vimrc and restart vim to make the changes take effect... There is a lot of friction in it.

# The possible solution
We already have dynamic marks with harpoon... So, why not use similar idea here. A user can configure the project commands dynamically with the help of provided API functions. Also, he can use some quick_command_menu kind of thing for visually changing the configured commands for project. When we have to send the command to terminal we send it by fetching the command from json (maybe named as harpoon_commands.json).

Now, users have the power of dynamically configuring the project commands that means tests and builds are so easy and so configurable now without messing our dotfiles.

Additional problem of having preconfigured commands in vimrc is the fact that most of people just maintain  their dotfiles as git repo and many people do some temporary projects and temporary project configuration in vimrc is just a pain in ass...

Any chance of adding this feature.... If you want I can work on it and submit the PR. I hope it would highly increase the usability of the plugin

ThePrimeagen:
I would be glad to try it out! I haven't thought too hard about the command portion. I am not convinced it's great as it is, and I think it needs work. But at the same time, I am unsure if I want to use harpoon to run commands.

I would definitely be glad to review the pr, give it a try out, if it works well and doesn't introduce any bugs I'd even be willing to merge it for you.

asbjornhaland:
You can use your `vimrc` to define the shortcuts to run commands. Then you can set the commands in your project specific config (`:h exrc`). Will this solution work - what do you guys think?

RishabhRD:
@asbjornhaland I have some problems with this approach. 

First thing is the exrc is now probably going to gitignore of project... And I don't like it... VS code guys already pollute it in some ways or other and just think you want to make PR to a project and maintainers use some other text editors. Are they really going to accept the PR that makes exrc to gitignore just for us because we use some other editor... Some of them do... But This really gets hard in practice.

Second thing is it's not dynamic. That means  once defined in exrc if you want to change commands in workflow you either need to restart vim or resource the exrc. In practice we can do it... But it's already a lot of friction in it... We just want smooth development with least keystrokes possible(yes we want coconut oil)

[...]

PR #97, opened 2021-08-29
harpoon dynamic commands for projects

RishabhRD:
As discussed in #96 , this PR adds dynamic_commands feature to harpoon.

I have done some additions in harpoon.term module to provide API to add and remove commands.
I have added another module named harpoon.cmd-ui, this module intends to provide quick menu features like marks quick menu.

Sample usecase video link: https://youtu.be/oI4kMpkADI4

### DIFF
--- a/lua/harpoon/cmd-ui.lua
+++ b/lua/harpoon/cmd-ui.lua
@@ -28,7 +28,7 @@ local function create_window()
     local width = config.width or 60
     local height = config.height or 10
     local borderchars = config.borderchars
-        or { "─", "│", "─", "│", "╭", "╮", "╯", "╰" }
+    or { "─", "│", "─", "│", "╭", "╮", "╯", "╰" }
     local bufnr = vim.api.nvim_create_buf(false, false)
 
     local Harpoon_cmd_win_id, win = popup.create(bufnr, {
@@ -42,9 +42,9 @@ local function create_window()
     })
 
     vim.api.nvim_win_set_option(
-        win.border.win_id,
-        "winhl",
-        "Normal:HarpoonBorder"
+    win.border.win_id,
+    "winhl",
+    "Normal:HarpoonBorder"
     )
 
     return {
@@ -54,7 +54,7 @@ local function create_window()
 end
 
 local function is_white_space(str)
-  return str:gsub("%s", "") == ""
+    return str:gsub("%s", "") == ""
 end
 
 local function get_menu_items()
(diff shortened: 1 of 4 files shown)


lua/harpoon/cmd-ui.lua
```diff
@@ -0,0 +1,148 @@
...
+    )
+
+    return {
+        bufnr = bufnr,
+        win_id = Harpoon_cmd_win_id,
+    }
+end
+
+local function is_white_space(str)
+    local white_space_chars = {
```
ThePrimeagen, line 57: we could probably just use \s and a regex here.  I KNOW... I SAID THAT... 

There should be a gsub call you can do, something like:

```
return #(str:gsub('%s', '')) == 0 -- not sure the exact api
```

lua/harpoon/cmd-ui.lua
```diff
@@ -0,0 +1,148 @@
...
+        if not is_white_space(line) then
+            table.insert(indices, line)
+        end
+    end
+
+    return indices
+end
+
+M.toggle_quick_menu = function()
+    log.trace("toggle_quick_menu()")
```
ThePrimeagen, line 87: add cmd-ui to the trace so we know which toggle quick menu.

lua/harpoon/cmd-ui.lua
```diff
@@ -0,0 +1,148 @@
+local harpoon = require("harpoon")
+local popup = require("popup")
+local log = require("harpoon.dev").log
+local term = require("harpoon.term")
+
+local M = {}
+
+Harpoon_cmd_win_id = nil
```
ThePrimeagen, line 8: I know I wrote this in the past, but I hate these global vars.  I feel like we shouldn't be using them :)

But we are currently, so I'll let it slide, but if I could get my way it would be to take these pop-up windows and just turn it into a class.

lua/harpoon/cmd-ui.lua
```diff
@@ -0,0 +1,148 @@
...
+    vim.cmd(
+        string.format(
+            "autocmd BufModifiedSet <buffer=%s> set nomodified",
+            Harpoon_cmd_bufh
+        )
+    )
+end
+
+M.on_menu_save = function()
+    log.trace("on_menu_save()")
```
ThePrimeagen, line 144: cmd-ui#on_menu_save()

lua/harpoon/cmd-ui.lua
```diff
@@ -0,0 +1,148 @@
...
+    end
+
+    vim.api.nvim_win_close(Harpoon_cmd_win_id, true)
+
+    Harpoon_cmd_win_id = nil
+    Harpoon_cmd_bufh = nil
+end
+
+local function create_window()
+    log.trace("_create_window()")
```
ThePrimeagen, line 26: I am a bit worried about so much duplicated logic.  I just don't want to create bugs that are hard to squuuuuoooosh

ThePrimeagen (approved):
I am going to merge this PR but I really do worry about the copied logic. 

I'll have to strongly consider reducing the two items used here, pop up windows + autocommands and "holed" lists into a class.  That way mutations can emit events easily without having to have this trixy logic spread around.

ThePrimeagen:
The only thing I ask is that you make the whitespace more simple.  For this case, the regex just works better since it will encompass all white space.  In general, the white space checking with regex's seems to be one of the only use cases I use for them.

lua/harpoon/cmd-ui.lua
RishabhRD: done the change

RishabhRD:
@ThePrimeagen Done with whitespace change and logger breadbutter. Didn't change the traces for private local functions because they would be called in context of one of public functions.

I can do the removal of code duplication by making classes for window and some free generic functions for holed lists... but that seems to be a refactor and would change the other files in project.... that doesn't seem me suitable for this PR. Am I right?

ThePrimeagen:
I'll merge this, try it out, and hopefully its awesome :)

ThePrimeagen:
@RishabhRD you have some lint :)

RishabhRD:
@ThePrimeagen I have updated luacheckrc for ignoring the global variables I have introduced as other global variables are also there as those would be refactored out in future versions. Also done some formatting things as formatting was failing.

RishabhRD:
I need to change formatting I guess. Actually formatting is quite messy with stylua. I don't know I tried to format document 3 times with given .stylua.toml but it said no changes required :sweat_smile: .

RishabhRD:
@ThePrimeagen I fixed the formatting and tested it using stylua also checked lint using luacheck... Please reivew it once

RishabhRD:
@ThePrimeagen any update on this PR?

ThePrimeagen:
Sorry this front end Masters presentation has been eating up all my time. Let me review this as soon as possible.

ThePrimeagen:
Uniroically I have merged this without review! I just want to make you happy! Will review later ^^

RishabhRD:
Thanks... I hope it would be useful for many :)

PR #98, opened 2021-09-04
Add mark functionality per git branch

aaronhallaert:
# The Problem
When jumping from one feature branch to another with git, I struggled with the fact that Harpoon only saved my marked files based on the current project. However, for every feature, I need to change code in different files corresponding to the feature (which I want to be marked).

# What changed
With this PR, I changed the behaviour to make sure every branch in the project has its own marks. Each project still has its general terminal commands which I can configure and use over the different branches.

# How to configure
Toggle this feature with a `mark_branch` property in harpoon settings.

Example:
```lua
require("harpoon").setup({
    mark_branch = true,
    projects = {
        ["$HOME/Developer/your-project"] = {
            term = {
                cmds = {
                    "docker exec -it web bash",
                }
            }
        }
    }
})
```

### DIFF
--- a/lua/harpoon/utils.lua
+++ b/lua/harpoon/utils.lua
@@ -4,17 +4,22 @@ local Job = require("plenary.job")
 
 local M = {}
 
-local project_key = vim.loop.cwd()
-M.project_key = project_key
-M.branch_key = vim.loop.cwd()
 M.data_path = data_path
 
-function M.mark_config_key()
-    return string.gsub(vim.loop.cwd() .. '-' .. vim.fn.system('git branch --show-current'), "\n", "")
+function M.project_key()
+    return vim.loop.cwd()
+end
+
+function M.branch_key()
+    return string.gsub(
+        vim.loop.cwd() .. "-" .. vim.fn.system("git branch --show-current"),
+        "\n",
+        ""
+    )
 end
 
 function M.normalize_path(item)
-    return Path:new(item):make_relative(M.branch_key)
+    return Path:new(item):make_relative(M.project_key())
 end
 
 function M.get_os_command_output(cmd, cwd)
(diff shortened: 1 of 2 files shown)


ThePrimeagen:
Off the bat I will say this has to be a configuration option, default current behavior. I already sort of have this solved with work trees.

lua/harpoon/utils.lua
```diff
@@ -1,10 +1,19 @@
 local Path = require("plenary.path")
 local data_path = vim.fn.stdpath("data")
+local project_key = vim.loop.cwd()
```
ThePrimeagen, line 3: Only feedback is lets make `project_key` a function and return current `vim.loop.cwd`.  If we do not, then we will get out of date harpoon menus when working with git worktrees.

btw, this change you want could also be done via git worktrees :)

ThePrimeagen (changes requested):

ThePrimeagen:
Just wanted to drop a note here.

If we get another request like this we will want to reduce all of these changes into a single call to get harpoon config path.  This will allow for users to provide their own functions as well.

ThePrimeagen:
@aaronhallaert address the one small problem (it would become a large problem if not :)) and then I'll merge.

lua/harpoon/utils.lua
aaronhallaert: Thanks for the feedback

I would like to use git worktrees, however, at work we mount our project folder to a docker container. Working with different folders for different branches would not allow me to jump branches as easily while keeping the same docker container running. I have no experience with git worktrees, so correct me if I'm wrong 😄

aaronhallaert:
@ThePrimeagen I implemented the change requests

harrisoncramer:
Hey Aaron I've tried running your forked version of Harpoon and noticed that when I've gotten it enabled, Neovim is significantly slowed down, to the extent that Telescope previews are timing out. Can post logfiles if you're interested. [Video](https://asciinema.org/a/A5dRvED72hv8tZp9SLN9LvQMV)

PR #100, opened 2021-10-08
Add convenience bindings for exiting quick menu

mikatpt:
## What changed?
Added bindings to quit the quick menu using either `q` or `<ESC>`.

## How was it tested?
Seems to work on my local as advertised - I'm afraid I don't know much about lua or vim testing :(

## Misc. comments
Small easy PR for now, will look into quit on buffer change/ focus lost later :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ You can simply edit this list as if it were a document in vim.  `:wq` to save
 the new edits or `:q` to ignore the edits.  There is to save upon call to
 toggle if you prefer that way.
 
+You can also exit the list with `q` or `<ESC>`, which will also save your edits!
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -78,6 +78,8 @@ You can simply edit this list as if it were a document in vim.  `:wq` to save
 the new edits or `:q` to ignore the edits.  There is to save upon call to
 toggle if you prefer that way.
 
+You can also exit the list with `q` or `<ESC>`, which will also save your edits!
```
ThePrimeagen, line 81: does it actually save?  It seems like this would just quit, which wont actually save

PR #102, opened 2021-10-20
Print file name and slot after calling add_file()

samodostal:
Hey Prime,
I have this problem with harpoon: When I call the `add_file()` command, I usually don't know in which slot the file will be placed. This usually results in calling `toggle_quick_menu()` right after to check the slot. This is especially a problem when harpoon has some files inside already. When I create a new project I can keep the slots in my head. Later on, I have to constantly check slots when adding new files.

I thought it might be a good idea to print to which slot was the current file placed in, after calling `add_file()`. Also, if the file is already placed in a slot, I would like it to print in which one.

I am not sure what the print message should look like (I am not that good at English :)). For now, this is how it looks like:
`print("File", buf_name, "added to slot", found_idx)`
`print("File", buf_name, "already exists in slot", buf_idx)`

Also, there could be an option in config when calling the `setup()`, to turn prints like this on.

If you don't like this idea, feel free to delete this PR. I will think of a different way to print the files just for me :)

### DIFF
--- a/lua/harpoon/mark.lua
+++ b/lua/harpoon/mark.lua
@@ -183,8 +183,10 @@ M.add_file = function(file_name_or_buf_id)
     local buf_name = get_buf_name(file_name_or_buf_id)
     log.trace("add_file():", buf_name)
 
-    if M.valid_index(M.get_index_of(buf_name)) then
+    local buf_idx = M.get_index_of(buf_name)
+    if M.valid_index(buf_idx) then
         -- we don't alter file layout.
+        print("File", buf_name, "already exists in slot", buf_idx)
         return
     end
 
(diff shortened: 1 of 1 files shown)


ThePrimeagen:
I like this idea.  But I think that we should try something a bit different.  

You can update your status line `H: %` where you replace `%` with a `harpoon#(i_forgot_the_function_call)`

If you really want this to work out, then we need to have an option that defaults to false and you can just have this feature only and set that option to true.

samodostal:
@ThePrimeagen, great idea! I like seeing it in the status line even more than printing it. This fixes my problem, and this PR can be closed. Thank you :)

@asbjornhaland Nice! I am using galaxyline, so I can use the function you created like this: 
```lua
require("galaxyline").section.left = {
    -- other galaxyline components
    {
     Harpoon = {
        provider = require("harpoon.mark").status,
        separator = '  ',
        highlight = { colors.white, colors.bg, 'bold' },
        separator_highlight = { 'NONE', colors.bg },
      }
    },
    -- other galaxyline components
}
```
Result (M1):

image

Thank you for help!

PR #106, opened 2021-10-27
remove popup.nvim as dependency

seblj:
This removes popup.nvim as a dependency. Not so long ago, popup.nvim was moved into plenary and is planned to be archived later.

https://github.com/nvim-lua/plenary.nvim/pull/209


ThePrimeagen:
awesome!  Ill try to get this merged here shortly!  :)

PR #112, opened 2021-11-04
feat: added excluded_filetypes option to avoid adding unwanted filety…

acicco:
### Problem

I accidentally added harpoon's buffer to the harpoon menu, this being an unwanted behaviour.

### Solution

After discussing, we decided to add an option to exclude filetypes that we want to prevent from being added to the harpoon's  list menu. 

So now it would look as follows:
```lua
require("harpoon").setup({
    global_settings = {
        save_on_toggle = false,
        save_on_change = true,
        enter_on_sendcmd = false,
        excluded_filetypes = {"harpoon"} -- harpoon being the default
    },
})
```
so every time the user decides to add a buffer with a filetype that is either harpoon or one that's on the list, the user should get an error message explaining that they can't add that specific filetype. As per user experience I was not sure if it should be an error or otherwise a warning.

Looking forward to hear your thoughts!


lua/harpoon/mark.lua
```diff
@@ -115,6 +115,18 @@ local function validate_buf_name(buf_name)
     end
 end
 
+local function filter_filetype()
+    local current_filetype = vim.bo.filetype
+    local excluded_filetypes = harpoon.get_global_settings().excluded_filetypes
+    for filetype = 1, #excluded_filetypes do
+        if(current_filetype == excluded_filetypes[filetype] or current_filetype == "harpoon") then
```
ThePrimeagen, line 122: lets move the harpoon check outside of the if statement that way you don't have to do it on every loop.  I mean, it doesn't matter, but it matters to me :)

ThePrimeagen (changes requested):

ThePrimeagen:
@acicco Fix the format / lint.

You could steal the `lint` task from refactoring.nvim if you want and make a make file.  Or the `pr-ready` task just no tests :)

lua/harpoon/mark.lua
acicco: That makes sense, I moved it out now with a different error message. Thanks Prime

ThePrimeagen (approved):

PR #114, opened 2021-11-07
Quick menu improvements

### LINKED ISSUE #113: Quick Menu Improvements
I've been playing around with this plugin since I watched [this masterpiece](https://youtu.be/Qnos8aApa9g) and I'm quite enjoying the workflow that `harpoon` provides to me. During my usage experience, I've been noticed some things that may be improved. Let me list them below.

## `'harpoon.ui'.toggle_quick_menu()`

It's really nice that the user can just edit the quick menu and then press `<Esc>` followed by `q` to exit or even an `<Enter>` to just open the entry of the selected line. However, I don't see the need to have to write the index of the entry before its path. I think the approach used in `'harpoon.cmd-ui'.toggle_quick_menu()` is a lot better. The quick menu for commands previously mentioned relies on the user to move lines around in order to switch the index of the commands in the list. It also has `number` set by default, which I think it's a better way to visualize the index of each entry. That said, I think that the quick menu for the `ui` module would be improved if behaved like the quick menu from `cmd-ui`.

## `'harpoon.cmd-ui'.toggle_quick_menu()`

This quick menu is almost perfect to me, but it lacks some essential features... The first one is that by pressing `<Enter>` in normal mode won't actually run the command in the selected line, which I think is a lot counterintuitive. The second one is that by pressing `q` in normal mode the menu won't close.

Really thanks for this plugin!

clbrunet:
Edited marks quick menu to use 'number' option instead of prefixing lines by indexes (which were no longer valid when reordering marks), in the same way as cmd quick menu.

Added 'q' and '\<ESC>' keymaps to cmd quick menu to close the window as in the marks quick menu.
And added \<CR> keymap to execute the command under the user's cursor (with a prompt to type the terminal index).

Closes #113

### DIFF
--- a/lua/harpoon/cmd-ui.lua
+++ b/lua/harpoon/cmd-ui.lua
@@ -109,14 +109,13 @@ M.toggle_quick_menu = function()
         ":lua require('harpoon.cmd-ui').toggle_quick_menu()<CR>",
         { silent = true }
     )
-    -- TODO: maybe vim.fn.input() can be used to implement some select_menu_item
-    -- vim.api.nvim_buf_set_keymap(
-    --     Harpoon_cmd_bufh,
-    --     "n",
-    --     "<CR>",
-    --     ":lua require('harpoon.cmd-ui').select_menu_item()<CR>",
-    --     {}
-    -- )
+    vim.api.nvim_buf_set_keymap(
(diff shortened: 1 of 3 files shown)


ThePrimeagen:
awesome.

Let me read

PR #115, opened 2021-11-09
fix: use <Cmd> instead of ':' in quick menus remaps

clbrunet:
Quick menus remaps now work even if a user uses a ':' remap.

It solves the problem raised by @rod-stuchi in #113.


ThePrimeagen:
I accept

Issue #99, opened 2021-09-09
Marks between directories


emr-arvig:
I have harpoon configured like Prime's dotfiles

The menu is always a new menu between directories

Sometimes I want to swap between a file in my tools dir and my utils dir within a greater project and it's a new harpoon menu for each directory.

Something simple I am missing?

I tried defining an empty project spec in config but this does not work either
```
require("harpoon").setup({
    global_settings = {
        save_on_toggle = true,
        save_on_change = true,
    },
    projects = {
        ["absolute path to super cool project"] = {}
    }
})
```

fritzlolpro:
Hi, got exactly same problem, after digging into code I found that projects is not about setting exact project, cuz there are lots of projects in runtime and every folder with mark counts as a project by default. 
![image](https://user-images.githubusercontent.com/10850158/142619761-14a6488a-682a-4d4d-ac1a-e664ccef3701.png)

I made a pull request to add the option you and me wanted -- having global marks.
Please take a look.

rogueinkamp:
Awesome this looks like it was merged? GREAT news if so. Thanks for doing the work on this!

fritzlolpro:
It wasn't merged, probably cuz my PR overlapsed with a huge code optimization + feature adding. 
I screwd up with upstream updating in my fork, so I made all changes anew, no PR yet, will make one soon.
I really love harpoon and think global marcks wouldnt harm

PR #122, opened 2021-11-19
Allow user to have global marks 

fritzlolpro:
Hello.
This PR should solve the usecase described in https://github.com/ThePrimeagen/harpoon/issues/99 and which also I encountered.
In the current implementation all marks are specific to the folder were the marked file is. 
My project have complex folder structure and I want to be able to navigate between files in all folders, so marks should be global.
In this PR I add the option to set global project were all marks will be saved
`global_project ="path/to/my/project"`
This approach is more convenient at least for me, and hope will be useful for others.

Without this option everything works as before.

btw I have no experience with lua, so pardon if I made some stupid things


fritzlolpro:
@ThePrimeagen could you please tell me if there is something I need to do in order to get this pr merged?

PR #124, opened 2021-11-22
telescope extension

### LINKED ISSUE #56: Telescope
https://github.com/brandoncc/telescope-harpoon.nvim

this should be moved into this plugin to be more easily seen and known to be used. should be easy as copying the telescope folder into the lua folder here

polarmutex:
closes #56

plan on trying to support all the current harpoon ui inside the telescope plugin

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -59,10 +59,7 @@ local function ensure_correct_config(config)
     log.trace("_ensure_correct_config()")
     local projects = config.projects
     if projects[vim.loop.cwd()] == nil then
-        log.debug(
-            "ensure_correct_config(): No config found for:",
-            vim.loop.cwd()
-        )
+        log.debug("ensure_correct_config(): No config found for:", vim.loop.cwd())
         projects[vim.loop.cwd()] = {
             mark = {
                 marks = {},
@@ -80,10 +77,7 @@ local function ensure_correct_config(config)
     end
 
     if proj.term == nil then
-        log.debug(
-            "ensure_correct_config(): No terminal commands found for",
-            vim.loop.cwd()
-        )
+        log.debug("ensure_correct_config(): No terminal commands found for", vim.loop.cwd())
         proj.term = { cmds = {} }
     end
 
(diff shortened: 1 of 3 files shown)


polarmutex:
do not  have a good solution to move marks to change the order,  any ideas would be appreciated

brandoncc:
Wow, this has so much more functionality than my extension. Nicely done!

mystilleef:
Works well. Way better than native UI. Kudos!

mystilleef:
@polarmutex can you set the `Preview` to show the exact position of the mark? Currently, the `Preview` only shows the top of the document.

polarmutex:
I am not sold on the current implementation of moving marks around, let me know if you have any better ideas

ThePrimeagen:
I don't have any ideas, but I am pretty stoked about this.

sQVe:
How do I actually use and register this to `telescope`. I'm a bit confused 🤔

polarmutex:
Probably should update the readme, I always forget that, I will work on it

But basically add extension to telescope and call "Telescope harpoon marks"

sQVe:
One thing I noted very quickly is that the entries seem to be truncated at a very short width. I would think that it should truncate at all.

image

sQVe:
Also, is there a way to override the defaults keymap here? I would like the invert `<Ctrl-n>` and `<Ctrl-p>`.

sQVe:
The entries seem to be truncated which I would assume it shouldn't be, since it breaks the ability to fuzzy search.

image

mystilleef:
@polarmutex the `Preview` is still not showing the location of the harpoon mark for me. It still only shows the beginning of the file. Am I missing something?

polarmutex:
i will take a look later today

sQVe:
It would be nice if we could make the confirm delete optional too, or that it supports deleting multiple entries. It's kinda a hassle to run <kbd>Ctrl+d</kbd> + <kbd>y</kbd> for multiple entries.

mystilleef:
Please test PR #130. It should fix the problem with the previewer not showing the exact location of marks.

PR #134, opened 2021-12-03
Create command.vim for Harpoon....

pynvimdev:
This will make it easy for begginer


matu3ba:
Please not. Users need to adjust the config anyway.

jamestrew:
I don't think there anything inherently wrong with having vim commands so I think it's fine.

matu3ba:
> I don't think there anything inherently wrong with having vim commands so I think it's fine.

At least not making them defaults would be nice.

jamestrew:
There's not defaults being set here. Just making it possible for people have to call harpoon functions using vim commands.

matu3ba:
This PR could now also use neovim lua commands, but I still dont see the point. The whole point of harpoon is to be fast, so navigating to file with a command feels sluggish.

PR #143, opened 2021-12-26
feat(bindings): add ctrl + c to close the quick menu

Taloud:
# What changed?
- Added bindings to quit the quick menu using <C-c>

# How was it tested?
Work on my local just have to try <C-c> on quick menu.

### DIFF
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -115,6 +115,13 @@ function M.toggle_quick_menu()
     vim.api.nvim_buf_set_keymap(
         Harpoon_bufh,
         "n",
+        "<C-c>",
+        "<Cmd>lua require('harpoon.ui').toggle_quick_menu()<CR>",
+        { silent = true }
+    )
+    vim.api.nvim_buf_set_keymap(
+        Harpoon_bufh,
+        "n",
         "<CR>",
         "<Cmd>lua require('harpoon.ui').select_menu_item()<CR>",
         {}
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ view all project marks with:
 ```lua
 :lua require("harpoon.ui").toggle_quick_menu()
 ```
-you can go up and down the list, enter, delete or reorder. `q` and `<ESC>` exit and save the menu
+you can go up and down the list, enter, delete or reorder. `q`, `<ESC>` and `<C-c>` exit and save the menu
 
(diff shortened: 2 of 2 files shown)


lpanebr:
looks good. I'll like to c-c.

ThePrimeagen:
@Taloud and @lpanebr qq

Have you tried any buffer specific remaps?  harpoon does come with its own buffer type and could possibly be the answer here without a change in code.

Thoughts?

lua/harpoon/ui.lua
```diff
@@ -112,6 +112,13 @@ function M.toggle_quick_menu()
         "<Cmd>lua require('harpoon.ui').toggle_quick_menu()<CR>",
         { silent = true }
     )
+    vim.api.nvim_buf_set_keymap(
```
ThePrimeagen, line 115: Before I merge this, we should consider perhaps a better way here.

To me, updating the README to show custom menu closing may be better.
If we merge this, I may want to make it so that its behind a flag to turn on.  It seems a bit gotcha for me

lpanebr:
> @Taloud and @lpanebr qq
> 
> Have you tried any buffer specific remaps? harpoon does come with its own buffer type and could possibly be the answer here without a change in code.
> 
> Thoughts?

That's a really good idea @ThePrimeagen !
I've added the following to my init.lua and it works perfectly:
```lua
vim.cmd [[
  autocmd FileType harpoon    nnoremap <buffer> <C-c> <cmd>lua require('harpoon.ui').toggle_quick_menu()<CR>
]]

```

Taloud:
> > @Taloud and @lpanebr qq
> > Have you tried any buffer specific remaps? harpoon does come with its own buffer type and could possibly be the answer here without a change in code.
> > Thoughts?
> 
> That's a really good idea @ThePrimeagen ! I've added the following to my init.lua and it works perfectly:
> 
> ```lua
> vim.cmd [[
>   autocmd FileType harpoon    nnoremap <buffer> <C-c> <cmd>lua require('harpoon.ui').toggle_quick_menu()<CR>
> ]]
> ```

Much better i think yes. Just add something like :

```lua
vim.cmd [[
  autocmd FileType harpoon nnoremap <buffer> <SHORTCUT> <cmd>lua require('harpoon.ui').toggle_quick_menu()<CR>
]]
```

to the readme and it's done :)

PR #146, opened 2022-01-08
Feature: Allow to provide a custom method to format the file path

johmsalas:
So far a DRAFT. I'd like to know how welcome is this feature.

It allows to format the showed filepaths, 2 examples:

image

image

I'm relatively new to Lua and Lua inside vim. Suggestions are welcome
Will continue giving this a try the next days before converting from DRAFT to complete PR. So far, deleting an item is failing

### DIFF
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -9,6 +9,10 @@ local M = {}
 Harpoon_win_id = nil
 Harpoon_bufh = nil
 
+local state = {
+    list_by_label = {}
+}
+
 -- We save before we close because we use the state of the buffer as the list
 -- of items.
 local function close_menu(force_save)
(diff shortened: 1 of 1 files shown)


lua/harpoon/ui.lua
```diff
@@ -63,13 +67,47 @@ local function get_menu_items()
...
+            else
+                table.insert(indices, line)
+            end
         end
     end
 
     return indices
 end
 
+-- This method will be moved to the readme as an example once the PR is not a DRAFT
```
johmsalas, line 82: The readme, or is there any suggested location for an example?

ThePrimeagen: The readme would be great.  How are you wanting to integrate this?  I would assume options, but I am unsure if my config makes it all through the program.  this was my first lua plugin for real, and it sucks.

johmsalas: Yeah, through the options, we can add an example in the readme, and provide builtin examples so that people do not need to repeat the same code everywhere, and if you think it is a good idea, people could still contribute their builtins (I think this would be similar to null-ls configs)


This is also my first Lua plugin (contribution). Having spent somedays with more readable paths was a good addition, I'm going to complete it and add automated tests later this week

johmsalas:
@ThePrimeagen can I please have your review on this?

lua/harpoon/ui.lua
```diff
@@ -81,15 +91,30 @@ function M.toggle_quick_menu()
...
+
     Harpoon_win_id = win_info.win_id
     Harpoon_bufh = win_info.bufnr
 
     for idx = 1, Marked.get_length() do
         local file = Marked.get_marked_file_name(idx)
-        if file == "" then
-            file = "(empty)"
+        local path = string.format("%s", file)
+        local label = formatter and formatter(file) or file
```
ThePrimeagen, line 107: thet thing that is tripping me up is why would there not be a formatter?

Can we just make a method that always ensure there is a formatter?

ThePrimeagen (changes requested):

lua/harpoon/ui.lua
ThePrimeagen: that is all my feedback.  I just think that we already have an identity function for the default.  Make it so then.

PR #150, opened 2022-01-23
Add ability to pass any tmux pane identifier

camgraff:
Currently, it is not possible to pass an existing tmux pane ID to harpoon. This can be useful to run a terminal command in a split pane in the same window, for example. 

This PR adds the ability to execute commands like 
```lua
lua require("harpoon.tmux").gotoTerminal("{down-of}")   -- focus the pane directly below
lua require("harpoon.tmux").sendCommand("%3", "ls")     -- send a command to the pane with id '%3'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ lua require("harpoon.tmux").sendCommand(1, "ls -La")    -- sends ls -La to tmux 
 lua require("harpoon.tmux").sendCommand(1, 1)           -- sends command 1 to tmux window 1
 ```
 
+`sendCommand` and `goToTerminal` also accept any valid [tmux pane identifier](https://man7.org/linux/man-pages/man1/tmux.1.html#COMMANDS).
+```lua
+lua require("harpoon.tmux").gotoTerminal("{down-of}")   -- focus the pane directly below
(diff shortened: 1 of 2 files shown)


ThePrimeagen:
ok ok ok ok

ThePrimeagen:
@camgraff why window vs pane?  I never use panes, can you give me a quick reason why this wont break things :)

camgraff:
> why window vs pane?

A window is just a container for panes. Currently, it forces creation of a new window (and hence a new pane) and [saves that window's pane id](https://github.com/ThePrimeagen/harpoon/pull/150/files#diff-97b743cd36d829534f5929d23cbd5e83c55c1cf8718feeb84955d2983ae89e7bL28). What if I want to send a command to a pane that already exists?

> can you give me a quick reason why this wont break things :)

The only change in functionality is when passing a string [which previously would throw an error](https://github.com/ThePrimeagen/harpoon/pull/150/files#diff-97b743cd36d829534f5929d23cbd5e83c55c1cf8718feeb84955d2983ae89e7bL74). Now, if you pass a string, it will assume it's a valid pane ID and if not, will return the error thrown by tmux. 100% backwards compat

PR #151, opened 2022-02-03
harpoon.ui: handle absolute paths

davvid:
If the filename is an absolute path then there is no need to
prepend the path with the current directory.

Detect absolute paths and open them as-is.

This fixes harpoon when editing files through symlinks
that point to other places on the filesystem.

### DIFF
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -164,7 +164,7 @@ function M.nav_file(id)
     end
 
     local mark = Marked.get_marked_file(idx)
-    local buf_id = vim.fn.bufnr(vim.fn.getcwd() .. "/" .. mark.filename, true)
+    local buf_id = vim.fn.bufnr(vim.api.cwd() .. "/" .. mark.filename, true)
     local set_row = not vim.api.nvim_buf_is_loaded(buf_id)
 
     vim.api.nvim_set_current_buf(buf_id)
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -164,7 +164,11 @@ function M.nav_file(id)
     end
 
     local mark = Marked.get_marked_file(idx)
-    local buf_id = vim.fn.bufnr(vim.api.cwd() .. "/" .. mark.filename, true)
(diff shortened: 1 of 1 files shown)


lua/harpoon/ui.lua
```diff
@@ -163,8 +163,14 @@ function M.nav_file(id)
         return
     end
 
+    local filename = nil
     local mark = Marked.get_marked_file(idx)
-    local buf_id = vim.fn.bufnr(vim.fn.getcwd() .. "/" .. mark.filename, true)
+    if mark.filename:sub(1, 1) == "/" then
+        filename = mark.filename
+    else
+        filename = vim.fn.getcwd() .. "/" .. mark.filename
```
disrupted, line 171: I think you could use `vim.loop.cwd()` which backed by libuv

disrupted (changes requested):

lua/harpoon/ui.lua
davvid: That same code is in the pre-image so I didn't change it, but if we think that's useful I can add an additional commit to swap over.

Yay / nay?

disrupted: yep, I know this is none of your "fault" so to say haha
simply noticed that this was still the old VimScript function, while newer APIs (and in this case especially because async) are preferred :)

davvid: Sweet, thanks for review ;-) I appreciate the free nvim education. I pushed in a new commit that swaps over to `vim.loop.cwd()`.

ThePrimeagen:
@davvid fix lint and i merge :)

davvid:
thanks, repushed. let's see if that does it

davvid:
I also rebased and rearranged the commit series so that the `vim.api.cwd()` swap is all by itself, before the bigger change instead of after it. much better

davvid:
crap sorry my stupid python brain. It's `~=` not `!=`. yeesh

davvid:
I believe this branch should be ready to re-lint and verify. I've been using this for a few weeks now; sorry for the lint issues.

PR #153, opened 2022-02-04
docs: add `mark_branch` blip in README

jamestrew:
Yup :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ feedback for harpoon (or me), make an issue!
 
 ## ⇁ The Problems:
 1. You're working on a codebase. medium, large, tiny, whatever. You find
-yourself frequenting a small set of files and you are tired of using a fuzzy finder, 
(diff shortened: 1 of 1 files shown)


harrisoncramer:
This feature only works for me when placing the `mark_branch = true` outside of the global settings block, but still within the setup table. Maybe I'm missing something. Like this:

```
		require("harpoon").setup({
			mark_branch = true,
			global_settings = {
				save_on_toggle = true,
			},
			menu = {
				width = 120,
			},
		})
```

jamestrew:
> This feature only works for me when placing the mark_branch = true outside of the global settings block, but still within the setup table.

I think you're right actually. I was fumbling around with the settings, I think I tried both and then kept only the `global_settings` one but I guess the settings get saved in `~/.local/share/nvim/harpoon.json` so it's persistent. I'm not the original author of the PR that introduced this feature but I guess my question is, should `mark_branch` be apart of `global_settings` or not?

jamestrew:
superseded by #161

PR #154, opened 2022-02-04
fix(tmux): fix tmux module to comply with latest changes to tmux API

pranavrao145:
It seems that `tmux` has recently made changes to the way some of their commands give output. Since Harpoon depends on some of those commands for the `harpoon.tmux` module, some changes were necessary to allow this to work.

### DIFF
--- a/lua/harpoon/tmux.lua
+++ b/lua/harpoon/tmux.lua
@@ -54,9 +54,9 @@ local function terminal_exists(window_id)
     -- This has to be done this way because tmux has-session does not give
     -- updated results
     for _, line in pairs(window_list) do
-        local window_info = utils.split_string(line, "@")[1]
+        local window_info = utils.split_string(line, "@")[2]
 
-        if string.find(window_info, window_id) then
+        if string.find(window_info, string.sub(window_id, 2)) then
             exists = true
         end
     end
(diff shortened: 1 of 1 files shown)


ThePrimeagen:
@pranavrao145 I'm just going to trust you our correct. I do not use harpoon and tmux.

svrana:
The change from select-pane to select-window breaks the goto terminal when using a pane identifier, like {right-of}. However, the initial select-pane change introduced by @camgraff  would break the case when the user is trying to move to a specific window and not a pane. I think we need to separate these two cases, using select-window or select-pane depending on the context to get this to work for both cases.

pranavrao145:
@svrana this is a good point, do you want to open a new issue on this?

svrana:
@pranavrao145 I've opened #158 and attached a patch, seems to work for both cases, but let me know if you have any problems.

pranavrao145:
Awesome, thanks @svrana, I'll test it out and let you know.

pranavrao145:
PS: Apologies, I think this one is on me; if I had looked at the commit history before committing this could have been avoided. Anyway, thanks for spotting that!

pranavrao145:
@svrana update, seems to work, thanks for the fix!

tg0h:
I am still having problems with this though ... the tmux command lua require('harpoon.tmux').gotoTerminal(2) always creates a new tmux window ... I will investigate further ...

PR #159, opened 2022-02-08
Fix tmux gotoTerminal for panes

### LINKED ISSUE #158: tmux gotoTerminal should support panes and windows
Pane support was added by @camgraff in 7cf2e20a411 which had the side-effect of breaking the gotoTerminal case when the user is using a tmux window.  @pranavrao145 fixed this in 627be003b, which broke the goto tmux pane support.

svrana:
gotoTerminal should work for both the pane and window case.

Resolves #158

### DIFF
--- a/lua/harpoon/tmux.lua
+++ b/lua/harpoon/tmux.lua
@@ -70,7 +70,10 @@ local function find_terminal(args)
     if type(args) == "string" then
         -- assume args is a valid tmux target identifier
         -- if invalid, the error returned by tmux will be thrown
-        return { window_id = args }
+        return {
+            window_id = args,
+            pane = true,
+        }
     end
 
     if type(args) == "number" then
@@ -118,7 +121,7 @@ function M.gotoTerminal(idx)
 
     local _, ret, stderr = utils.get_os_command_output({
         "tmux",
-        "select-window",
+        window_handle.pane and "select-pane" or "select-window",
         "-t",
         window_handle.window_id,
     }, vim.loop.cwd())


lua/harpoon/tmux.lua
```diff
@@ -115,10 +118,15 @@ end
...
+    local specifier = "select-window";
+
+    if window_handle.pane then
+        specifier = "select-pane"
+    end
 
     local _, ret, stderr = utils.get_os_command_output({
         "tmux",
-        "select-window",
+        specifier,
```
disrupted, line 129: just FYI this could be done inline perhaps 
```lua
window_handle.pane and "select-pane" or "select-window"
```

svrana: Much nicer, thank you @disrupted.

svrana:
This is ready for another CI run.. Force pushed so maybe that notification was lost.

ThePrimeagen:
I don't do any of the tmux stuffs.  So I am not 100% i follow this logic.

I am worried because I see `pane = true`. should this be based on config?

svrana:
> I am worried because I see `pane = true`. should this be based on config?

That's what's happening. To configure tmux to work with windows, you specify a window index, i.e., an integer, like so:

```lua
lua require("harpoon.tmux").sendCommand(1, 'make')
```

However, if you want to this to work with a pane, you specify its position with a string, i.e.

```lua
lua require('harpoon.term').sendCommand('{bottom-right}', 'make')
```

tg0h:
What is the expected behaviour of go to terminal for panes if the pane does not exist, will it create a new pane?  

# current behaviour
- go to terminal for tmux panes
  - Currently, it does not create a new pane and fails silently,
- go to terminal for tmux windows
  - Creates a window if it does not exist

# expected behaviour
🤔

PR #161, opened 2022-02-10
refactor: improve git branch specific marks impl

jamestrew:
A few changes from #98 

1. Use `git rev-parse --abbrev-ref HEAD` over `git branch --show-current` as it's more commonly available on old git versions.
2. Reduced the number of calls to the aforementioned command to improve performance
3. Make `mark_branch` a global_setting
4. Add `mark_branch` under global_settings in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ feedback for harpoon (or me), make an issue!
 
 ## ⇁ The Problems:
 1. You're working on a codebase. medium, large, tiny, whatever. You find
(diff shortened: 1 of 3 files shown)


aaronhallaert:
Great work! 
I wrote the original code for the git branch marks (which were btw my first lines of lua), but this PR really improves the performance. 🚀
@harrisoncramer noticed me about preview timeouts in Telescope because of my changes. I believe this PR fixes this issue.

ThePrimeagen:
Many TYs

PR #172, opened 2022-03-28
Feat: Allow user to switch back to the main neovim window

yonikosiner:
Hiya, I have been using harpoon for a while, and I love the fact that I can quickly to switch to my tmux windows, but I was never able to switch back to neovim as smoothly. So I have wrote this little script that allows you to switch back to my neovim session.

**If you have more then one window open running neovim at the moment it will not work, I'm working on a fix for that.**


ThePrimeagen:
this is fantastic.  I think that i might incorporate this

PR #192, opened 2022-06-21
Try to switch back to window named "nvim" only if at least one exists

kmiasko:
Now if i try to use switch-back-to-nvim when window named "nvim" doesn't exist there is ugly message, that tmux command returned with exit status 1. I've added small check to verify if there is one in current session.


ThePrimeagen:

PR #193, opened 2022-06-27
Allow user to have global marks

fritzlolpro:
Hello.
I made similar PR before, but looks like it overlaped with some big changes. So I made everything anew.
https://github.com/ThePrimeagen/harpoon/pull/122 <--- old pr

This pull request will allow user to add global marks, the user will be able to decide where the root folder should reside. The most common case (my case) is to set the root folder as a git root folder like
       ` global_project = vim.fn[  's:find_git_root'  ]()`
where
```
  function! s:find_git_root()
    return system('git rev-parse --show-toplevel 2> /dev/null')[:-2]
  endfunction
```

There is also an feature request issue for this functionality https://github.com/ThePrimeagen/harpoon/issues/99

Hope this feature will be usefull.

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -109,11 +109,8 @@ local function ensure_correct_config(config)
             marks[idx] = mark
         end
 
-        if config.global_project then
-            marks[idx].filename = mark.filename
-        else
-            marks[idx].filename = utils.normalize_path(mark.filename)
-        end
+        marks[idx].filename = utils.normalize_path(mark.filename, config.global_project)
+
     end
 
     return config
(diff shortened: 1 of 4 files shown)


ThePrimeagen:

fritzlolpro:
My code is terrible, I accept it, im not expirenced with lua and nvim plugins. So yes, it could break something as there is also lots of interconnected parts in existing code imo and I've tested my changes only manually on very limited number of cases.
Atm I'm using my fork of Harpoon and it seems to be working for me.
Still can't give 100% that something wouldn't break, so dunno what to do honestly.
Are there some lua testing suites so I culd try to test all the stuff?

PR #194, opened 2022-06-30
Added functions to allow scripting of whether to open a new terminal

matt-gardner:
I wanted to run a command that queries whether there is already a terminal (and if not, open a split window, create a new terminal there, and return to the original window, but that happens in my code, not harpoon).  These functions allow me to do that.


ThePrimeagen:

PR #210, opened 2022-09-14
feat(term): add global setting create_with

aenichols:


aenichols:
@ThePrimeagen Any chance of getting a review?

PR #212, opened 2022-09-22
fix(ui): avoids concatenating dubble cwd for the poor windows users.

OlofMoriya:


matu3ba:
This should probably now use the `vim.fs` API available since neovim 0.8.

PR #215, opened 2022-10-01
Implement move up for telescope extension

MaximilianLloyd:
I saw this todo in the code so i implemented it.

PS: The formatting on lines 53-55 came from running `stylua lua/telescope/_extensions/marks.lua --config-path=.stylua.toml` so i assume it's alright?


MaximilianLloyd:
I didn't change the utils file. So it's strange that it fails.

PR #217, opened 2022-10-13
feat: title of the window changes as buffers are changed

rosaelton:
Before adding this line, when we started changing buffers with harpoon, the window title kept being the same as the first opened file. With this line, the title changes dynamically as we change buffers.


matu3ba:
This is not optional and can be distracting, if one is quickly and often changing buffers. To me it would more sense to show the `cwd`, but even then that is not job of the plugin and plugins could even have conflicting decisions or worse autocommands permanently changing the title.

PR #224, opened 2022-10-28
Fix item before plenary.paths

mikaellpc4:
I was having a problem with harpoon on windows, somethimes for some reason vim.api.nvim_buf_get_name() gives the path of the directory with the partition letter undercase like "c:\\" or put / instead of \\, this things just breaks the plenary normalizer

I created this "solution" and it's working fine for me at the moment

If anyone can improve this code please doit, i'm realy new to Lua so this maybe can be something ugly or bad i don't no

### DIFF
--- a/lua/harpoon/utils.lua
+++ b/lua/harpoon/utils.lua
@@ -27,8 +27,15 @@ function M.branch_key()
     end
 end
 
+local function fixItem(item)
+    local upperItem = item:gsub('^%l:', string.upper)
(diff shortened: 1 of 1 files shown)


matu3ba:
This should be fixed upstream in neovim and not hacked around in harpoon.

ThePrimeagen:
@mikaellpc4 have you made an issue with neovim?  I think that would be a better place for this fix.  teej_dv, and the neovim team, are super fast to respond.

PR #226, opened 2022-11-02
fix: do not force absolute file paths

marcelbeumer:
When forcing absolute filepaths buffer names will get that absolute filepath too. To me it seems that vim.fn.bufadd supports relative paths just fine.

### DIFF
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -173,11 +173,7 @@ function M.nav_file(id)
     end
 
     local mark = Marked.get_marked_file(idx)
-    local filename = mark.filename
-    if filename:sub(1, 1) ~= "/" then
-        filename = vim.loop.cwd() .. "/" .. mark.filename
-    end
-    local buf_id = get_or_create_buffer(filename)
+    local buf_id = get_or_create_buffer(mark.filename)
     local set_row = not vim.api.nvim_buf_is_loaded(buf_id)
 
     vim.api.nvim_set_current_buf(buf_id)


matu3ba:
Did you test, if this breaks the following workflow?

- 1. Assume `cwd` and 1 buffer with `keybinding` for loading is stored in json
- 2. nvim `cwd`
- 3. `:cd somepath`
- 4. press `keybiniding` to load file

One uses absolute paths to force paths to be always correct, if `cwd` of neovim changes.

marcelbeumer:
sorry @matu3ba I don't understand exactly what flow you are suggesting.

Harpoon should work always in context of a cwd. You could put up to discussion if harpoon should rather store absolute filepaths instead of relatives (which is does not), and then force to display relative if possible, but this PR just works with what harpoon was already providing. 

My understanding of the code is that M.nav_file is run within context of the current pwd.

matu3ba:
> Harpoon should work always in context of a cwd.

Is this documented anywhere and I missed it? Otherwise, it would be nice to add something like "harpoon stores paths based on cwd, so`:cd` to `../...../someotherproject` are not supported" into the manual and README.

> if harpoon should rather store absolute filepaths instead of relatives

From my understanding, if the path is above the project, for example `$HOME/myotherdir` vs the stored `$HOME/projects/harpoonfiles`, then it will store absolute paths. Underlying reason is that `plenary.path:new(vim.api.nvim_buf_get_name(0)):make_relative())` stores them like that, but I need to recheck what harpoon uses to resolve relative paths (the alternative is to use realpath, but that is not available on apple/mac).

I have this entry in my harpoon.json, where I was setting some marks from a unittest dir and searched via copy-paste into  `:e path`. The same can be applicable to vendored git repos etc.
```json
{
....
"$HOME/projects/somepath/unittests": {
  "mark": {
    "marks": [
      {
        "col": 16,
        "row": 69,
        "filename": "$HOME/projects/baredir/gitworktree/src/lib/somefile.h"
      },
      {
        "col": 0,
        "row": 82,
        "filename": "$HOME/projects/baredir/gitworktree/src/lib/somefile.cpp"
      },
      {
        "col": 0,
        "row": 26,
        "filename": "$HOME/projects/baredir/gitworktree/tests/src/lib/someUT.cpp"
      }
    ]
  },
  "term": {
    "cmds": []
  }
},
...
}
```

matu3ba:
This looks almost identical to this PR: #238.

ThePrimeagen:
I want to redo this whole repo

So I am hesitant to merge anything to avoid breaking

When Christmas is over, I am going to think about the rewrite

matu3ba:
This PR should be closed as #238 was merged.

PR #233, opened 2022-11-21
Performance improvements 

rsynnest:
I'm hoping to use this PR to make a few performance improvements.

The first commit is a minor performance improvement to speed up buffer switching. The culprit is [this call](https://github.com/ThePrimeagen/harpoon/blob/f4aff5bf9b512f5a85fe20eb1dcf4a87e512d971/lua/harpoon/init.lua#L20) to `store_offset()` which is setup as a `BufLeave` autocmd. It loops through all harpoon marks to find a match for the current buffer, and stores the cursor's row + column position for that mark, all before leaving the buffer. The more marks we add, the more marks we have to loop through when buffer switching. This commit moves the `get_mark_config()` call outside our loop which marginally improves buffer switching performance.

Will try to dig into other optimizations like a faster JSON parser or anything else I can find.

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -144,7 +144,7 @@ end
 
 local function read_config(local_config)
     log.trace("_read_config():", local_config)
-    return vim.fn.json_decode(Path:new(local_config):read())
+    return vim.json.decode(Path:new(local_config):read())
 end
 
 -- 1. saved.  Where do we save?
--- a/lua/harpoon/mark.lua
+++ b/lua/harpoon/mark.lua
@@ -136,7 +136,7 @@ local function filter_filetype()
     end
 end
 
-function M.get_index_of(item)
+function M.get_index_of(item, marks)
     log.trace("get_index_of():", item)
     if item == nil then
         log.error(
@@ -150,8 +150,10 @@ function M.get_index_of(item)
 
     if type(item) == "string" then
         local relative_item = utils.normalize_path(item)
-        local marks = harpoon.get_mark_config().marks
-        for idx = 1, M.get_length() do
+        if marks == nil then
+            marks = harpoon.get_mark_config().marks
+        end
+        for idx = 1, M.get_length(marks) do
             if marks[idx] and marks[idx].filename == relative_item then
                 return idx
             end
@@ -191,13 +193,13 @@ function M.status(bufnr)
     return ""
 end
 
-function M.valid_index(idx)
+function M.valid_index(idx, marks)
     log.trace("valid_index():", idx)
     if idx == nil then
         return false
     end
 
-    local file_name = M.get_marked_file_name(idx)
+    local file_name = M.get_marked_file_name(idx, marks)
     return file_name ~= nil and file_name ~= ""
 end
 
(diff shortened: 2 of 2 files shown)


matu3ba:
> faster JSON parser

https://github.com/xpol/lua-rapidjson or you could take a bigger effort and complete the simdjson lua bindings: https://github.com/simdjson/simdjson#bindings-and-ports-of-simdjson

Another alternative is to use the Zig bindings for simdjson + build system and smack the bindings on top. Not sure which and if an existing lua binding generator for Zig works, but writing them by hand sounds simpler (to me) than completing the simdjson lua bindings.

rsynnest:
@matu3ba I haven't done any profiling, any tools you recommend? I'm new to Lua.

lua-simdjson looks complete enough for decoding which I suspect to be the main bottleneck. It also has [similar performance to lua-rapidjson and lua-csjon](https://github.com/FourierTransformer/lua-simdjson#benchmarks) so any of those might work.

After reading up on [shared libs in Lua](http://lua-users.org/wiki/BuildingModules), it sounds like I'll need to either have end users compile the module at plugin installation, or distribute shared libraries for all platforms and use a bit of bash or Lua to load the right one. Could maybe use Zig for cross compilation, I've heard it's good for that.

disrupted:
doesn't the builtin `vim.json.decode` already use cjson?
Perhaps it would be better to consider replacing the JSON storage entirely with something that has better scalability.

matu3ba:
> Perhaps it would be better to consider replacing the JSON storage entirely with something that has better scalability.

Do you have a suggestion, ideally with benchmark justification for the used data, which remains human inspectable or always backwards compatible + very simple for converting to string?
The only thing, which is widely used and comes to my mind would be json-rpc, but I dont know how much perf difference that would be, as we have mostly string data.

matu3ba:
> I haven't done any profiling, any tools you recommend?

Accurate results should use https://blast.hk/moonloader/luajit/ext_profiler.html as we use luajit as lua runtime. Take a look at this list https://github.com/nvim-lua/wishlist/issues/4. For completeness, there is also the inbuild one (startup profiler) and packer has one.

disrupted:
why does it have to be human-readable? JSON has a pretty significant overhead over binary data, because each record contains the whole schema.
I was thinking more about the likes of SQLite or similar. https://www.sqlite.org/whentouse.html

rsynnest:
Tbh I'm not 100% sure of how Lua memory works in Neovim (is there some type of main loop? can we persist objects in memory). But I think the big win will be using an in-memory Lua table for all the data for a given project. Whether that's JSON or sqlite probably doesn't matter too much, some research needed there. I imagine on startup the plugin could load project-specific data from disk into the Lua table (let's call it memdb), and it would use memdb as the source of truth for the whole session. Any changes made to memdb would get pushed to disk asynchronously with something like `vim.schedule()`.

If you want to support multiple separate vim sessions for the same project then it gets a little fucky. The simplest option is probably to do a diff+merge before every disk write. Maybe a beacon file would be faster to indicate a merge is needed (maybe use Shada for this?). If you can assume a single user doing one thing at a time, might have simpler options. 

[neoclip](https://github.com/AckslD/nvim-neoclip.lua) might be a good reference for this style of data management? doesn't use vim.schedule but it's simple to read through at least.

matu3ba:
@rsynnest I think this is more reasonable to discuss in a dedicated issue. It would likely help to have some accurate latency numbers on usage and provide users an option to measure and report them.
Or describe some specific use cases, ideally with simple to reproduce examples, where you think its definitely too slow.

Adding a script to create the config file for users (to create comparable situations) could also help to narrow down things.

PR #235, opened 2022-11-24
Open a harpoon file from the quickmenu in a vertical or horizontal split

psaikido:
or a new tab.


psaikido:
I would add that lots of folks use splits and tabs due to it being natively easy to yank and paste.

ThePrimeagen:
i ackshually will merge this one, very great idea

ThePrimeagen:
i am positive there is a better way to do this, and there is lots of duplicate code in there, but for now, i'll allow it :)

PR #245, opened 2023-01-17
Add clear marks command

alxekb:
Add readme how to clear marks list

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,10 @@ you mark files you want to revisit later on
 :lua require("harpoon.mark").add_file()
 ```
 
+```lua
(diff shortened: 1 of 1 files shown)


alxekb:
@ThePrimeagen should we mention this in the readme?

evanasse:
There is a `clear_all()` function for that (https://github.com/ThePrimeagen/harpoon/blob/f7040fd0c44e7a4010369136547de5604b9c22a1/lua/harpoon/mark.lua#L292), so it might be helpful to add it to the README.

olejech:
I found how to clear all marks. Thanks

PR #246, opened 2023-01-20
Remove empty `[No Name]` buffer when navigating to file.

Khachig:
I have an annoying experience every time I navigate to one of my harpoon files right after launching vim that roughly goes as follows:

- Launch vim (opens empty buffer)
image
- Navigate to first harpoon file (opens file, leaves empty buffer loaded)
image
- Switch back to empty unnamed buffer
- Delete empty unnamed buffer
image

This change addresses this issue by adding a few lines of code to check whether the current  buffer before navigating is an empty unnamed buffer and deletes it after navigation:

- Launch vim (opens empty buffer)
image
- Navigate to first harpoon file (opens file, deletes empty buffer)
image


ThePrimeagen:
nice catch

PR #260, opened 2023-02-23
Fix buf with style

ThePrimeagen:


ThePrimeagen:
i am merging this even though there is a style failure.

there is a production bug and i don't know what is wrong with the styler.

```

PR #262, opened 2023-02-24
allow users to get terminal and buffer id for customization.

matu3ba:


matu3ba:
Example usage:
```lua
M.setCursorToBottom = function(harpoon_term_nr)
  local term_handle = harp_term.getBufferTerminalId(harpoon_term_nr)
  local buf_id = term_handle["buf_id"]
  local term_id = term_handle["term_id"]
  local windows = vim.fn.win_findbuf(buf_id)
  if #windows ~= 1 then
    return
  end
  local crowcount = vim.api.nvim_buf_line_count(0)
  vim.api.nvim_win_set_cursor(windows[1], { crowcount, 0 })
end
```

Unfortunately a (neo)vim bug prevents reliable usage.

matu3ba:
If it is desired to ~~scrap apart things piece by piece~~ rebuild things as library, then this could pave the way to deprecate things piece by piece, give users copy-pasteable config examples to keep their workflows going and make things nice.

Though not sure, if its worth the churn.

PR #264, opened 2023-03-03
Bind numbers to marks in quick menu

willothy:
Quickly jump to an entry by typing its number from the quick menu!

### DIFF
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -103,6 +103,8 @@ function M.toggle_quick_menu()
             file = "(empty)"
         end
         contents[idx] = string.format("%s", file)
+        vim.api.nvim_buf_set_keymap(Harpoon_bufh, "n", string.format("%d", idx),
+            "<Cmd>lua require('harpoon.ui').nav_file(" .. idx .. ")<CR>", { silent = true })
     end
 
     vim.api.nvim_win_set_option(Harpoon_win_id, "number", true)


pockata:
Hey, this is an interesting addition, however it overrides the default ability to slice n dice the quick menu with simple movements like `3j dd 2k p` and so on.

I see a lot of creative PRs in the repo that are waiting to be merged, but probably won't be, because they serve a niche workflow that may or may not benefit the rest of the Harpoon user base.

Perhaps a better approach would be to add the functionality as an autocmd for the Harpoon filetype so anyone can add it to their config if they so desire. We can then gather these enhancements in a Recipes section in the README or even a Recipes Wiki page in the repo so that anyone can cherry-pick exactly what they need.

@ThePrimeagen you on board with this?

willothy:
> Hey, this is an interesting addition, however it overrides the default ability to slice n dice the quick menu with simple movements like `3j dd 2k p` and so on.

Haha, that is a good point. Not sure how I didn't think of that, I guess I just mainly use Harpoon with a small number of buffers that I switch between quickly, so I rarely move more than 1 or 2 lines in the quick menu. I'll refactor this into an autocmd and submit a new PR whenever there's a recipes section/page :)

ThePrimeagen:
so i generally agree with the sentiment of @pockata 

it would be too one off, but i do recommend creating remaps only for harpoon buffers then you could have normal maps to move quickly between positions :)

(but i always recommend vim motions where applicable)

asbjornhaland:
See my comment in another issue: https://github.com/ThePrimeagen/harpoon/issues/283#issuecomment-1540948309

willothy:
> (but i always recommend vim motions where applicable)

Yep! After using Harpoon with some larger projects, I switched back to <kbd>num</kbd>-<kbd>j/k</kbd> as not being able to jump beyond 0-9 without spamming keys became a bit annoying.

PR #277, opened 2023-04-08
Use fugitive for faster branch name resolution

pockata:
There's a noticeable slowdown when opening the quick menu or switching to a file when `mark_branch` is enabled. It's present even on modern hardware with a fast ssd. Barely noticeable but still enough to destroy that creamy-smooth harpoon feel.

I've identified the cause of the slowdown to be the call to the `git` binary when getting the branch name.

Given the ubiquity of tpope's `vim-fugitive`, I've used `FugitiveHead` to get the branch name. It doesn't rely on the `git` binary so it's noticeably faster.  It's not a hard dependency, so when fugitive is not available, it falls back to the current method - calling `git rev-parse`.

### DIFF
--- a/lua/harpoon/utils.lua
+++ b/lua/harpoon/utils.lua
@@ -11,14 +11,25 @@ function M.project_key()
 end
 
 function M.branch_key()
-    -- `git branch --show-current` requires Git v2.22.0+ so going with more
-    -- widely available command
-    local branch = M.get_os_command_output({
-        "git",
-        "rev-parse",
-        "--abbrev-ref",
-        "HEAD",
-    })[1]
+    local branch
+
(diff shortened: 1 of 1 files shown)


pockata:
It seems the stylua checks have failed on chunks of code not related to this PR. I can format them and include them in this PR if requested.

ThePrimeagen:
i approve of this, but i need you to fix the stylings :)

ThePrimeagen:
oh snap, i just read your above comment, then yes, i may have to merge this in.  tehre is definitely some styllua issues

RafaelZasas:
PackerSync let me know that this had some potential breaking changes. Should this soydev relax and let Prime take the wheel?

PR #279, opened 2023-04-10
Auto add quick mapping to menu files based on index

carlossgv:
Generate automatic quick mappings to files while Harpoon menu is open.

For example, if file is in index 2 you can go to file by pressing 2.


carlossgv:
Just checked your comments, makes a lot of sense. 

Closing the PR, thanks for the feedback!

PR #289, opened 2023-05-17
fix: redraw tabline whenever marks change

milkias17:
Right now when you add a file to harpoon, the tabline doesn't update and add your new file.

- Now `tabline.setup` registers a callback on the changed event of `harpoon.marks`.
- I couldn't `require('mark')` inside the `tabline.setup` because it caused a circular import. This is why I required `harpoon` inside each function in `harpoon.marks` so that I could call `require('mark')` without causing a circular import
- I couldn't come up with a better solution.

### DIFF
--- a/lua/harpoon/mark.lua
+++ b/lua/harpoon/mark.lua
@@ -1,4 +1,3 @@
-local harpoon = require("harpoon")
 local utils = require("harpoon.utils")
 local log = require("harpoon.dev").log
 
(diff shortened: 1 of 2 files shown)


otavioschwanck:
I did the same with less code here (plus a fix):

https://github.com/ThePrimeagen/harpoon/pull/290

milkias17:
Don't you think its bad practice to check global settings inside of an emitter function. I presumed the whole point of the event system implemented was so that other functions like the tabline could register callbacks to run on a certain event and the event emitter doesn't need to know internal details.

In any case, thanks a lot for the tabline, it has made my workflow close to perfect. Really amazing job, thanks a lot.

PR #293, opened 2023-05-25
Added notifications when marking files

dormunis:


davvid:
Related-to: #317 

Unlike #317, this PR has one additional difference which might (or might not) tip the scales for someone -- this PR differentiates between when a buffer is added and when a buffer is not added (because it's already marked).

In my opinion this PR would also benefit from using a callback function (as described in #317) but with an additional `already_exists` boolean to differentiate the two scenarios.

I don't know if that would be reason enough to support this feature, though, so please don't waste your time updating PRs based on a random cat's opinion unless someone else chimes in. I'm just a random harpoon user.

PR #301, opened 2023-07-18
feat(telescope): Improve telescope results

### LINKED ISSUE #160: Smart text truncation for file names
I work on a monorepo so my file paths are extremely long and will wrap over multiple lines. Often all I care about is the folders two levels above the current file path. I would like to see my harpooned files each on a single line with the text left aligned and truncated as to allow me to just see the 2+ folders above a file.

Telescope has this built in with path_display={"smart"}

### LINKED ISSUE #225: Support Telescope path_display option
Telescope is able to shorten the displayed file paths using the `path_display` option. It would be nice if the harpoon extension would respect this setting (or offer something similar).

### LINKED ISSUE #237: Formatting Configurations for Filenames
I'm currently working in a monorepo, and I kinda have to eyeball my current files in both Harpoon via telescope, and the quick menu. It's pretty annoying if components are similarly named. I did some quick searching over lunch about possibly utilizing telescope's `smart` and `truncate` options for path_display, but didn't find if it was possible to apply those options on custom made finders. Harpoon is a great tool, but this is a real pain point in my workflow. 

image

geralfonso:
- fix truncated entries
- Add telescope path_display and icon support
/
Fixes #237 
Fixes #225 
Fixes #160 

## Before
Screenshot 2023-07-18 at 5 17 45 PM

## After
Screenshot 2023-07-18 at 5 11 37 PM


alicanerdogan:
This is brilliant. I wonder if we can ask @ThePrimeagen to have a look?

PR #303, opened 2023-08-07
feat(ui): dynamic window width

pmwals09:
Allows setting a function as the menu width, to enable dynamic floating window widths.

I frequently work with terminal splits in Tmux, and would run into the following situation:
1. Open NeoVim in full terminal, do some work, mark files w/ Harpoon
2. Split terminal to run a command on the work I just did, and leave it running
3. Go back to NeoVim and toggle Harpoon's window to change files

On #3, the width is set on init, so Harpoon's menu is still set according to the full terminal size and goes off the NeoVim buffer. By making adjustments to allow setting a function as the width instead, we can avoid this issue. The usual method of setting the value explicitly still works.

Example usage that will set the window width to be 80% of the window width or 60, whichever is greater:
```lua
require("harpoon").setup {
  menu = {
    width = function()
      return math.max(60, math.floor(vim.api.nvim_win_get_width(0) * 0.8))
    end
  }
}
```


lua/harpoon/cmd-ui.lua
```diff
@@ -26,7 +26,12 @@ end
 local function create_window()
     log.trace("_create_window()")
     local config = harpoon.get_menu_config()
-    local width = config.width or 60
+    local width
+    if type(config.width) == "function" then
+        width  = config.width() or 60
+    else
+        width  = config.width or 60
+    end
```
EdenEast, line 34: This can be changed to have fallback conditionals.

```suggestion
    local width = type(config.width) == "function" and config.width()
        or config.width
        or 60
```

lua/harpoon/ui.lua
```diff
@@ -28,7 +28,12 @@ end
 local function create_window()
     log.trace("_create_window()")
     local config = harpoon.get_menu_config()
-    local width = config.width or 60
+    local width
+    if type(config.width) == "function" then
+        width = config.width() or 60
+    else
+        width = config.width or 60
+    end
```
EdenEast, line 36: Same comment as the cmd-ui line above:
```suggestion
    local width = type(config.width) == "function" and config.width()
        or config.width
        or 60
```

lua/harpoon/utils.lua
```diff
@@ -74,4 +74,18 @@ function M.is_white_space(str)
...
+        for orig_key, orig_value in next, orig, nil do
+            copy[M.deepcopy(orig_key)] = M.deepcopy(orig_value)
+        end
+        setmetatable(copy, M.deepcopy(getmetatable(orig)))
+    else
+        copy = orig
+    end
+    return copy
+end
+
```
EdenEast, line 90: Deep copy function is not required after local temp changes.

lua/harpoon/init.lua
```diff
@@ -171,7 +170,12 @@ function M.save()
     M.refresh_projects_b4update()
 
     log.trace("save(): Saving cache config to", cache_config)
-    Path:new(cache_config):write(vim.fn.json_encode(HarpoonConfig), "w")
+
+    local to_save = utils.deepcopy(HarpoonConfig)
+    if type(to_save.menu.width) == "function" then
+        to_save.menu.width = nil
+    end
+    Path:new(cache_config):write(vim.fn.json_encode(to_save), "w")
```
EdenEast, line 178: A deep copy is not required. This can be achieved with a local temp variable. 

```suggestion
    local menu = HarpoonConfig.menu
    HarpoonConfig.menu = nil
    Path:new(cache_config):write(vim.fn.json_encode(HarpoonConfig), "w")
    HarpoonConfig.menu = menu
```

EdenEast (commented):
Thanks for making this PR. I wrote a similar PR before seeing that this PR already existed. Instead of making another PR with similar changes I have reviewed this PR instead.

pmwals09:
@EdenEast This was great feedback. On the newer side to Lua so this is very helpful. Thanks!

PR #305, opened 2023-08-12
feat(tabline): added format function to export

Mikilio:
The function can be exported via M.format(opts). I also got rid of the global function, so my LSP is happy. The actual tabline string has been edited as well to work more smoothly with lualine.

### DIFF
--- a/.luarc.json
+++ b/.luarc.json
@@ -1,0 +1,3 @@
+{
+    "completion.autoRequire": false
+}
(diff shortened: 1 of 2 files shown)


Mikilio:
What needs to be done here?

ThePrimeagen:
Master is frozen. All development is going into harpoon2

It will replace out master within the next few months

There is a tabline task if you wish to take a look at it.

PR #316, opened 2023-10-01
Run commands in another tmux window and return back

kareem-abdul:
Hi,
I saw this [video](https://www.youtube.com/watch?v=hJzqEAf2U4I) on @ThePrimeagen channel in which he was trying to run a command in another TMUX window which after exiting come's back automatically to his previous TMUX window (neovim/vim).

Even though that video was specific for cht.sh I had other cases where I wanted the same behavior. And I also wanted it to work with the existing harpoon command buffer.
i.e. you open the buffer with a key-bind select the command with j and k press enter, run the command in a new tmux window, wait for some input, and return back to the previous window.

### DIFF
--- a/lua/harpoon/cmd-ui.lua
+++ b/lua/harpoon/cmd-ui.lua
@@ -3,6 +3,7 @@ local popup = require("plenary.popup")
 local utils = require("harpoon.utils")
 local log = require("harpoon.dev").log
 local term = require("harpoon.term")
+local tmux = require("harpoon.tmux")
 
 local M = {}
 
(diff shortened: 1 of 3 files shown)


ThePrimeagen:
i am going to make this possible by Harpoon 2, i don't want it in harpoon 1

Harpoon 2 is going to allow for more complicated operations, and i'll consider this one (that way you can make harpoon-tmux, instead of harpoon having TMUX integrated)

kareem-abdul:
> i am going to make this possible by Harpoon 2, i don't want it in harpoon 1
> 
> Harpoon 2 is going to allow for more complicated operations, and i'll consider this one (that way you can make harpoon-tmux, instead of harpoon having TMUX integrated)

yup makes sense.

PR #317, opened 2023-10-02
feature: notification when adding new mark

kevintraver:
new opt-in feature to notify when adding harpoon mark.

Can be set with `notify_on_add` global setting

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,9 @@ global_settings = {
     tabline = false,
     tabline_prefix = "   ",
     tabline_suffix = "   ",
+
+    -- notify when adding a mark
+    notify_on_add = false,
 }
 ```
 
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -213,6 +213,7 @@ function M.setup(config)
             ["tabline"] = false,
             ["tabline_suffix"] = "   ",
             ["tabline_prefix"] = "   ",
+            ["notify_on_add"] = false,
         },
     }, expand_dir(c_config), expand_dir(u_config), expand_dir(config))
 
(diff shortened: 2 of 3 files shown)


davvid:
Here's my opinion as a minimalist ~ hopefully it's at least slightly thoughtful. My general stance is that it's better to avoid adding features unless they're somewhat widely beneficial or provide novel utility.

The predefined behavior provided by this feature makes me wonder, "what if I want to use a different message?," "what if I want to change the log level?," and so on.

I don't think the answer is to add more configuration.

If this were added then it seems like a more general approach would be to allow the user to provide a callback function that receives the buffer name as an argument so that the user can provide their own custom behavior rather calling vim.notify for them.

But, that then begs the question: why do even need to provide a callback mechanism in the first place?

Users already have to configure harpoon keybindings in their config and it's pretty easy to bind to a function that calls vim.notify before or after calling harpoon.mark.add_file() today. Thus, it doesn't seem this is core functionality that the plugin should provide given how easy it is for users to call vim.notify themselves in their own custom keybindings.

The only small roadblock is that `get_buf_name()` isn't a public function. This means that the user would need to call `require('harpoon.utils').normalize_path(vim.api.nvim_buf_get_name(0))` to get the buffer name, but that's not really that hard to do either IMO.

That's more or less why I don't think this feature should be added ~ this behavior seems like it'd be better off in your init.lua rather than in harpoon. That's just my opinion, though.

kevintraver:
Thanks for the response, it makes total sense! I agree it's good to weigh the benefit of adding more code to a plugin.

kevintraver:
I'll go ahead and close this out for now.

PR #320, opened 2023-10-08
Fix inheritance of 'number' option

name-snrl:
There are a few more changes here than I originally envisioned. The problem I wanted to solve is that when you open a file in a new window (e.g. `C-v`), it inherits the `number` option from the harpoon window. The easiest way to solve this problem is to replace the deprecated (in the next release, I'm building nvim from master) `nvim_win_set_option` with `nvim_set_option_value` (released in 7.0).

I then noticed that `nvim_buf_set_option` was also deprecated, so I replaced all `nvim_buf_set_option` and `nvim_win_set_option` with `nvim_set_option_value` and changed the minimum required neovim version in the README.

Finally, I run stylua.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ to go to the files you want.
 
 
 ## ⇁ Installation
-* neovim 0.5.0+ required
+* neovim 0.7.0+ required
 * install using your favorite plugin manager (`vim-plug` in this example)
 ```vim
 Plug 'nvim-lua/plenary.nvim' " don't forget to add this one if you don't have it yet!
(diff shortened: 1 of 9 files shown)


name-snrl:
cc @ThePrimeagen

ThePrimeagen:
we are fixing all these issues in Harpoon2 which is almost done (i just have to port over the UI part of harpoon 1 and add autocmds)

it has more flexible api.  i will update this issue when harpoon2 is ready to be used

name-snrl:
Wow, nice. Thank you

PR #321, opened 2023-10-09
fix: events no longer can be comma separated

deathmaz:
After https://github.com/neovim/neovim/pull/25523 the comma separated events are not valid any more and will throw error


willothy:
@ThePrimeagen please consider merging this, Harpoon is completely unusable on latest nightly without this fix.

PR #331, opened 2023-11-10
feat: menu title option

sand4rt:


ThePrimeagen:
We are in the process of writing harpoon 2, so any merges like this will not be taken care of

Be freezing master

PR #349, opened 2023-12-01
add self.config default item in remove function

zachlankton:
Before this change I could not remove the current buffer from the harpoon list using the `list():remove()`

image


ThePrimeagen:
very interesting!!! removing the current buffer by calling empty remove.  I love the idea.  i am a bit worried about calling `add` in the `remove`  function

i may break this later :)  but i will replace it with a specific method.

or i may rename `add` to `create_list_item` which is probably the right move here

Issue #381, opened 2023-12-06
harpoon2 open in split/tab mappings


name-snrl:
Hello, are there any plans to add these mappings to harpoon 2? Or now I will have to setup it myself through autocmds or any other way?

https://github.com/ThePrimeagen/harpoon/blob/867e212ac153e793f95b316d1731f3ca1894625e/lua/harpoon/init.lua#L32-L54

name-snrl:
I tried adding mapping to `buffer.lua`, like:

```lua
    vim.api.nvim_buf_set_keymap(
        bufnr,
        "n",
        "<C-V>",
        "<Cmd>lua require('harpoon').ui:select_menu_item({ vsplit = true })<CR>",
        {}
    )
```

but it doesn't work properly, #382 fixes that

ThePrimeagen:
so opening in split / tab totally works!  but not from auto-commands.  this is an interesting issue

so #382 is not the way i wish to go about it, but we should be able to extend the config in such a way that when selecting there is a "default" option OR when opening a toggle menu we can pass an `option` for the config select action to be used.

if you have any ideas, glad to see a PR, else i'll try to get to this as one of the many tasks of getting harpoon 2 off the ground

ThePrimeagen:
If this is fixed could you close this down?

ThePrimeagen:
it is fixed based on the PR i just closed

name-snrl:
> it is fixed based on the PR i just closed

This issue requested for a mapping to open files from harpoon UI to (v)split/tab, why are you closing it? There is still no mapping to open files in split or even the ability to add them to harpoon, I don't understand why you think the problem is solved.

psaikido:
This doesn't look like it has been fixed to me. I put the fix in for original harpoon with this commit (https://github.com/ThePrimeagen/harpoon/commit/edf29292da0481ed434a104262592e891c4c028f). I see no way to open a file in a split or tab from harpoon2.

Issue #379, opened 2023-12-06
Harpoon2 when pressing q/esc to exit UI, it moves to a marked item


whoop-t:
Whichever item in the UI you have highlighted, then press q/esc(to exit ui), it will move to that item as if you pressed enter on it.
Neovim v0.9.0
Test on both Linux(popos) and Macos

Config:
```lua
  "ThePrimeagen/harpoon",
  lazy = false,
  branch = "harpoon2",
  config = function()
    local harpoon = require "harpoon"
    harpoon:setup()

    vim.keymap.set("n", "<leader>a", function() harpoon:list():append() end)
    vim.keymap.set("n", "<C-e>", function() harpoon.ui:toggle_quick_menu(harpoon:list()) end)
    vim.keymap.set("n", "<C-j>", function() harpoon:list():select(1) end)
    vim.keymap.set("n", "<C-k>", function() harpoon:list():select(2) end)
    vim.keymap.set("n", "<C-l>", function() harpoon:list():select(3) end)
    vim.keymap.set("n", "<C-;>", function() harpoon:list():select(4) end)
  end,
  dependencies = { "nvim-lua/plenary.nvim" },

```

theKnightsOfRohan:
I've also been getting this error.

whoop-t:
Confirming this is fixed with #392 
tyty, Closing

PR #382, opened 2023-12-06
Remove `ui:toggle_quick_menu` from BufLeave autocmd

name-snrl:
I don't know what that line is for, but removing it fixes some issues.

### DIFF
--- a/lua/harpoon/buffer.lua
+++ b/lua/harpoon/buffer.lua
@@ -110,7 +110,6 @@ function M.setup_autocmds_and_keymaps(bufnr)
         pattern = "__harpoon*",
         callback = function()
             require("harpoon").logger:log("toggle by BufLeave")
-            require("harpoon").ui:toggle_quick_menu()
         end,
     })
 end


ThePrimeagen:
what issue are you having that you need this?  the reason why this exists is if you leave the a buffer/navigate the harpoon window should close.  that is this purpose.  but you may be on to something i haven't thought of!

name-snrl:
> what issue are you having that you need this?

#379

to reproduce:
- open nvim and add a few marks
- exit
- open nvim again, without any file
- open and close menu
- you will be dropped to one of the mark

as for #381, when I tried to add mapping with `vsplit` option to `buffer.lua`, just as hardcode for tests
[harpoon.patch](https://github.com/ThePrimeagen/harpoon/files/13587803/harpoon.log)

Regarding #381, when I tried to add a [mapping with the `vsplit` option in `buffer.lua`](https://github.com/ThePrimeagen/harpoon/files/13587803/harpoon.log), just as a hardcode to test the performance, it didn't work. But when I applied the patch from this PR, the mapping started working correctly

name-snrl:
#386 is the right solution

PR #386, opened 2023-12-06
fix nav issues with q and escape

### LINKED ISSUE #379: Harpoon2 when pressing q/esc to exit UI, it moves to a marked item

tris203:
This resolves #379 

I dont like the fix for escape, as it wouldn't save changes to the list
but in the logs, it would always come through a <CR> rather than an escape, not sure why.

Although, thinking about it, is Esc more of a cancel action then a save action
Open to thoughts

### DIFF
--- a/lua/harpoon/buffer.lua
+++ b/lua/harpoon/buffer.lua
@@ -24,6 +24,10 @@ end
 function M.run_toggle_command(key)
     local harpoon = require("harpoon")
     harpoon.logger:log("toggle by keymap '" .. key .. "'")
+    if key == 'q' or key == '<Esc>' then
+        harpoon.ui:close_menu()
+        return
+    end
     harpoon.ui:select_menu_item()
 end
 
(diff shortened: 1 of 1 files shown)


tris203:
updated

name-snrl:
Wait, shouldn't we just replaces `'<ESC>'` with 'Esc' in rhl of escape mapping, because `nvim_buf_set_keymap()` replaces `<ESC>` with termcode, and the command just doesn't execute? I mean:

```patch
-         "<Cmd>lua require('harpoon.buffer').run_toggle_command('<ESC>')<CR>",
+         "<Cmd>lua require('harpoon.buffer').run_toggle_command('Esc')<CR>",
```

_keycodes are replaced as usual._

https://github.com/neovim/neovim/blob/8bb5089974e14f043a0b0c17e4d38dbd0b9a9ab2/runtime/doc/api.txt#L1508

ThePrimeagen:
Sorry the other PR was a bit more comprehensive and was the way I wanted

Issue #341, opened 2023-11-29
[Feature request] ability to disable cycling


pmielle:
hey,
I think it would be useful to be able to disable cycling when using nav_next() and nav_prev(). 

For example: if I have 4 harpoons and I am currently at the 4th one, I would like `nav_next()` to do nothing. Same thing for `nav_prev()` on nav(1)

I know it would make sense for some of my workflows where I tend to go deeper and deeper into a codebase, setting checkpoints along the way

thanks for the hard work!

catgoose:
> hey,
> 
> I think it would be useful to be able to disable cycling when using nav_next() and nav_prev(). 
> 
> 
> 
> For example: if I have 4 harpoons and I am currently at the 4th one, I would like `nav_next()` to do nothing. Same thing for `nav_prev()` on nav(1)
> 
> 
> 
> I know it would make sense for some of my workflows where I tend to go deeper and deeper into a codebase, setting checkpoints along the way
> 
> 
> 
> thanks for the hard work!

What about using tag stack instead?

pmielle:
Thank you for your answer! 
disclaimer: I am not familiar with tag stacks

I like harpoon because I manually set which file I want to keep in my save list and which one I don't.
If I understand correctly, tab stack keeps track of every jump I make unconditionally.. which is not exactly what I would like to have. And on top of that, I would guess that tab stacks do not persist when closing and opening back vim, but maybe I'm wrong here

but I get what you mean: tab stacks solve the "go deeper and deeper into a codebase"-thing I described. Still, I tend to prefer harpoon for the reasons I mentioned above

ThePrimeagen:
@pmielle i am not really understanding this

how would you envision this working?  would there be a harpoon function you would call to go into this "don't change state" mode?  i am just unsure if i want to support this but i want to understand it.

pmielle:
oh no no no. I would just like to have a setting to prevent harpoon to loop back to the beginning of the list when hitting `nav_next()` on the last item. Sorry if I'm not making myself clear (I'm french..)

ThePrimeagen:
oh yes! i see i see.  this makes perfect sense.  that seems like a reason able option

i am now releasing Harpoon2 as the new harpoon (interface change, but TONS of new power + less features (which gives more options for more features))

pmielle:
ooh awesome! looking forward to it. thanks for your feedback + for your content in general!

ThePrimeagen:
[...]

PR #387, opened 2023-12-06
Add feature to disable/enable wrapping on next/prev when cycling through harpoon lists.

mwishoff:
Implements the ask on #341. Allows for people to configure the usage of harpoon_wrap in harpoon:setup(). By default harpoon_wrap is True which keeps the wrapping behavior that already existed.

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -74,6 +74,7 @@ function M.get_default_config()
         },
 
         default = {
+            harpoon_wrap = true,
             --- select_with_nill allows for a list to call select even if the provided item is nil
             select_with_nil = false,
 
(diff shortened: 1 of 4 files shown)


ThePrimeagen:
Let's not call this harpoon_wrap, but instead ui_nav_wrap, and default value true

mwishoff:
Updated the PR. should I update the rest of the formatting that's failing the check to get the format checker to pass?

ThePrimeagen:
I'm not enforcing it at a local level yet, this is still exploratory branch. Things are going very good though

ThePrimeagen:
I think we should be able to test this. Would you mind creating a failing test? Even if it's just eq(true, false) with a name for me to remember to create it

jjiangweilan:
This feature has been moved. look at https://github.com/ThePrimeagen/harpoon/pull/461#issuecomment-1879332220

Issue #385, opened 2023-12-06
Opening items in the list from the UI throws an error


fn3x:
### How to reproduce
1. Add 2 or more files to harpoon list
2. Open one of them using harpoon keymap or UI selection
3. Open another using harpoon UI
After that every opening from the UI throws an error

Config:
```
{
  menu = {
    width = vim.api.nvim_win_get_width(0) - 4,
  },
  settings = {
    save_on_toggle = true,
  },
}
```

### Error trace
```
E5108: Error executing lua ...Data/Local/nvim-data/lazy/harpoon/lua/harpoon/config.lua:128: BufLeave Autocommand for "__harpoon*": Vim(append):Error executing lua callback: ...Data/Local/nvim-data/lazy/harpoon/lua/harpoon/buffer.lua:120: Expected Lua number
stack traceback:
	[C]: in function 'nvim_buf_get_lines'
	...Data/Local/nvim-data/lazy/harpoon/lua/harpoon/buffer.lua:120: in function 'get_contents'
	.../AppData/Local/nvim-data/lazy/harpoon/lua/harpoon/ui.lua:153: in function 'save'
	.../AppData/Local/nvim-data/lazy/harpoon/lua/harpoon/ui.lua:114: in function 'toggle_quick_menu'
	...Data/Local/nvim-data/lazy/harpoon/lua/harpoon/buffer.lua:113: in function <...Data/Local/nvim-data/lazy/harpoon/lua/harpoon/buffer.lua:111>
	[C]: in function 'nvim_set_current_buf'
	...Data/Local/nvim-data/lazy/harpoon/lua/harpoon/config.lua:128: in function 'select'
	...ppData/Local/nvim-data/lazy/harpoon/lua/harpoon/list.lua:168: in function 'select'
	.../AppData/Local/nvim-data/lazy/harpoon/lua/harpoon/ui.lua:148: in function 'select_menu_item'
	...Data/Local/nvim-data/lazy/harpoon/lua/harpoon/buffer.lua:21: in function 'run_select_command'
	[string ":lua"]:1: in main chunk
stack traceback:
	[C]: in function 'nvim_set_current_buf'
	...Data/Local/nvim-data/lazy/harpoon/lua/harpoon/config.lua:128: in function 'select'
	...ppData/Local/nvim-data/lazy/harpoon/lua/harpoon/list.lua:168: in function 'select'
[...]

PR #391, opened 2023-12-07
fix: verify if self.bufnr is not nil before save

daltondiaz:
I reproduce the error and see that when you don't change the ui Harpoon try to save but buffer number don't exist, so this cause an error.

Issue: https://github.com/ThePrimeagen/harpoon/issues/385


daltondiaz:
Better approach: https://github.com/ThePrimeagen/harpoon/pull/392

PR #394, opened 2023-12-09
feat: highlight the current file in the menu

### LINKED ISSUE #384: Toggling harpoon list should highlight current file
If you have 5 files in the harpoon list and you have file 4 selected, it would be nice if when toggling harpoon list open, file 4 is highlighted instead of the first one.

This would make it easy to open list, go down one or two items then select that one.

If the buffer in the current window is not in harpoon list, then highlight the first item.

willothy:
Adds highlighting to the current file in the menu, and removes the old `matchadd` method of doing so which did not work. 

fixes #384

### DIFF
--- a/lua/harpoon/buffer.lua
+++ b/lua/harpoon/buffer.lua
@@ -32,18 +32,6 @@ end
 ---strings back into the ui.  it feels gross and it puts odd coupling
 ---@param bufnr number
 function M.setup_autocmds_and_keymaps(bufnr)
-    local curr_file = vim.api.nvim_buf_get_name(0)
-    local cmd = string.format(
-        "autocmd Filetype harpoon "
-            .. "let path = '%s' | call clearmatches() | "
-            -- move the cursor to the line containing the current filename
-            .. "call search('\\V'.path.'\\$') | "
-            -- add a hl group to that line
-            .. "call matchadd('HarpoonCurrentFile', '\\V'.path.'\\$')",
-        curr_file:gsub("\\", "\\\\")
-    )
-    vim.cmd(cmd)
-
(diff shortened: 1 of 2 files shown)


tris203:
I was also looking at a PR for this.

The only issue I can flag, is that in config.lua the path is normalised a different way using a normalise_path function.

i think that for consistency, either that function should be moved to utils and used in both scripts. or the config file should be modified to use the plenary.path:absolute() method too.

willothy:
> I was also looking at a PR for this.
> 
> The only issue I can flag, is that in config.lua the path is normalised a different way using a normalise_path function.
> 
> i think that for consistency, either that function should be moved to utils and used in both scripts. or the config file should be modified to use the plenary.path:absolute() method too.

I agree that it would be ideal for them to be the same, not sure why I did that haha. If you want to work on this I'll close this PR and you can move forward with yours, otherwise I'll make that change :)

tris203:
Happy to do either.

We should probably get input from @ThePrimeagen on which would be preferred. I think to use the : absolute() class in both

I also think that potentially this should be behind an option to either start at the top of the list or follow the file down the list

willothy:
> Happy to do either.

For sure! You seem to have more ideas about what you want from this, so why don't you do it. 

> We should probably get input from @ThePrimeagen on which would be preferred. I think to use the : absolute() class in both

I think the choice to use normalized paths was intentional so I'm not sure. Either way there's going to be some conversion from absolute to normalized though, since the list items shouldn't be displayed as absolute paths. Agreed with getting input.

> I also think that potentially this should be behind an option to either start at the top of the list or follow the file down the list

I like that idea, would definitely use it.

PR #395, opened 2023-12-09
feat: introduce move event

willothy:
Anotha one

Adds a `move` event so we can have similar tracking to v1's `harpoon.on("changed", ...)`. I use the event to trigger reordering of buffers in my tabline according to their order in harpoon (falling back to standard buffer order for non-marked bufs).


tris203:
Is Move ambiguous? Would Reorder be clearer?

willothy:
> Is Move ambiguous? Would Reorder be clearer?

Yeah I think that's a good idea

PR #199, opened 2022-08-23
feat: Make current file easily noticeable

pockata:
This PR contains two changes that revolve around a single purpose - to make the currently opened file easily noticeable in the quick menu.

The first one is non-breaking and it adds a highlight group (`HarpoonCurrentFile`, not linked by default) on the line that contains the currently opened file.

The second is somewhat breaking and moves the cursor to the line that contains the currently opened file when you open the quick menu.

If the current buffer is not Harpooned, they do nothing.

Here's a preview to help explain the features.

[Peek 2022-08-23 22-40.webm](https://user-images.githubusercontent.com/543561/186251943-0021f2f3-fe6c-48a7-9277-94dff3dc55e9.webm)

(Here, `HarpoonCurrentFile` is linked to Search - `:highlight link HarpoonCurrentFile Search`)

Rationale: In the couple of months that I've been using Harpoon, I often found myself editing the harpoon list around the currently opened file (move it to the top, or remove all other files, etc), so making it more easily noticeable feels natural.

PR #397, opened 2023-12-09
Auto highlight selected

tris203:
#396 but to the right branch
Adds request in #384

### DIFF
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -3,7 +3,6 @@ local Buffer = require("harpoon.buffer")
 local Logger = require("harpoon.logger")
 local utils = require("harpoon.utils")
 
-
 ---@class HarpoonUI
 ---@field win_id number
 ---@field border_win_id number
@@ -110,9 +109,8 @@ end
 ---@param list? HarpoonList
 function HarpoonUI:toggle_quick_menu(list)
     local currentFileName = utils.normalize_path(
-        vim.api.nvim_buf_get_name(
-            vim.api.nvim_get_current_buf()
-        ))
+        vim.api.nvim_buf_get_name(vim.api.nvim_get_current_buf())
+    )
 
     if list == nil or self.win_id ~= nil then
         Logger:log("ui#toggle_quick_menu#closing", list and list.name)
@@ -134,7 +132,14 @@ function HarpoonUI:toggle_quick_menu(list)
     vim.api.nvim_buf_set_lines(self.bufnr, 0, -1, false, contents)
     for idx, line in ipairs(contents) do
         if line == currentFileName then
-            vim.api.nvim_buf_add_highlight(self.bufnr, -1, "HarpoonCurrentFile", idx - 1, 0, -1)
+            vim.api.nvim_buf_add_highlight(
+                self.bufnr,
+                -1,
+                "HarpoonCurrentFile",
+                idx - 1,
+                0,
+                -1
+            )
             if self.settings.select_current then
                 vim.api.nvim_win_set_cursor(self.win_id, { idx, 0 })
             end
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -1,6 +1,8 @@
 local popup = require("plenary").popup
 local Buffer = require("harpoon.buffer")
 local Logger = require("harpoon.logger")
+local utils = require("harpoon.utils")
+
 
 ---@class HarpoonUI
 ---@field win_id number
@@ -107,6 +109,13 @@ end
 
 ---@param list? HarpoonList
 function HarpoonUI:toggle_quick_menu(list)
+    local currentFileName = nil
+    if self.settings.select_current then
+        currentFileName = utils.normalize_path(
+            vim.api.nvim_buf_get_name(
+                vim.api.nvim_get_current_buf()
+            ))
+    end
 
     if list == nil or self.win_id ~= nil then
         Logger:log("ui#toggle_quick_menu#closing", list and list.name)
@@ -126,6 +135,13 @@ function HarpoonUI:toggle_quick_menu(list)
 
     local contents = self.active_list:display()
     vim.api.nvim_buf_set_lines(self.bufnr, 0, -1, false, contents)
+    if self.settings.select_current then
+        for idx, line in ipairs(contents) do
+            if line == currentFileName then
+                vim.api.nvim_win_set_cursor(self.win_id, { idx, 0 })
+            end
+        end
+    end
 end
 
 ---@param options? any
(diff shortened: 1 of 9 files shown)


tris203:
@willothy thoughts?

lua/harpoon/ui.lua
```diff
@@ -126,6 +135,13 @@ function HarpoonUI:toggle_quick_menu(list)
 
     local contents = self.active_list:display()
     vim.api.nvim_buf_set_lines(self.bufnr, 0, -1, false, contents)
+    if self.settings.select_current then
+        for idx, line in ipairs(contents) do
+            if line == currentFileName then
+                vim.api.nvim_win_set_cursor(self.win_id, { idx, 0 })
+            end
```
willothy, line 142: You are setting the cursor, but not applying highlights. I would suggest to move the if statement inside the loop:

```lua
for idx, line in ipairs(contents) do
  if line == currentFileName  then
    vim.api.nvim_buf_add_highlight(...)
    if self.settings.select_current then
      vim.api.nvim_win_set_cursor(...)
    end
    break -- also make sure to break after you find the line
  end
end
```

lua/harpoon/utils.lua
```diff
@@ -1,8 +1,14 @@
...
 
+function M.normalize_path(buf_name, root)
+    return Path:new(buf_name):make_relative(root)
+end
+
 function M.trim(str)
-   return str:gsub("^%s+", ""):gsub("%s+$", "")
+    return str:gsub("^%s+", ""):gsub("%s+$", "")
 end
+
```
willothy, line 11: You should run this thorough `stylua`

willothy (commented):
Looks good to me so far, but you're not setting the highlight, just the cursor :)

lua/harpoon/utils.lua
tris203: I have

lua/harpoon/ui.lua
tris203: Done

tris203 (commented):
Ready for review

lua/harpoon/utils.lua
```diff
@@ -1,8 +1,14 @@
+local Path = require("plenary.path")
 local M = {}
 
+function M.normalize_path(buf_name, root)
+    return Path:new(buf_name):make_relative(root)
+end
+
 function M.trim(str)
-   return str:gsub("^%s+", ""):gsub("%s+$", "")
+    return str:gsub("^%s+", ""):gsub("%s+$", "")
```
willothy, line 9: lgtm! Except for this singular extra space here lol

tris203: 4 space indenting, thats where its meant to be

willothy: LOL idk how I didn't see that, my bad.

tris203:
I assume, given the direction that this will need adjusting to pull the logic out of UI

I would guess the best way is to make toggle_quick_menu() take a second optional param of CurrentItem which is used for the highlighting/focus logic

Then a function passed back from the API get_current() which contains the logic for determining the name of the current item

ThePrimeagen:
so i am going to be removing all highlighting from Harpoon and i am going to start emitting out events with win/buf id, and perhaps some extra state such that you can add whatever highlights you wish

i don't want harpoon having too much logic.  So i am "likely" not going to accept any PRs that add logic but instead adding hooks

willothy:
Actually this does raise an important point (even if the highlighting is removed from this PR). Not sure how I didn't think of this before, but what if the list items aren't paths? I think that if there is going to be a way to determine the current item, it needs to be a method on the List Config so that different lists can modify the default behavior. It also needs to be optional because for some lists (commands, for example) there is no "current" item, so it would not make sense to show one.

ThePrimeagen:
@willothy there is equals and display

From the UI we call display, else we call equals

willothy:
Yes, sorry I should've been more clear. I mean that we're always checking `nvim_get_current_buf` at the moment so even if we do use `equals` and `display`, we're still only able to check against lists of buffers because the current thing in `equals(list_item, current)` will always be the name of the current buffer.

ThePrimeagen:
yes, these things can be tricky

to me its an information problem.  perhaps we could provide "from_buffer" or some option to let you know whence you came

willothy:
> to me its an information problem. perhaps we could provide "from_buffer" or some option to let you know whence you came

Yes, that's perfect. A more generic way to check "is this buffer related to this item" than checking the name.

pockata:
I was just about to rewrite in Lua that SOB that I PR'd in #199 to drop the `vim.cmd` and `string.format` calls, but I noticed this PR :) 

I just wanted to point out that with `vim.api.nvim_buf_add_highlight` when you `dd` the highlighted entry and paste it again it will lose the highlighting. That's why the better approach is to use `vim.fn.matchadd` (perhaps `vim.fn.matchaddpos` too, but I haven't used it) which preserves the highlighting after an edit (technically reapplies it automatically when a string matches the pattern). 

`HarpoonCurrentFile` isn't widely used because you have to dig into the source code to find it. That's bad marketing on my behalf. If it was present in the README, for example, I believe this would have been a fan favorite feature. An incredibly sensible default of `highlight link HarpoonCurrentFile String` is an amazing way to spot the current item in the list `blazingly fast`, especially if you're a bit dyslexic. Having the highlighting in a hook which you grab from a "Recipes" wiki page is also a good way to go, but having it as a built-in default will provide a much better experience for new and experienced users alike.

I suggest you add the highlight command in your config, test drive it for a couple of days and then remove it to see if you miss it :)

ThePrimeagen:
Coming back to this i think i want to go a different direction.

i would like to think about how to make this a extension not something inside of harpoon.  If you have an idea for an event to emit where you could hook in and do this work yourself, please do.  I would have no problems with the harpoon-colors plugin doing something for me.  I just don't want to maintain that :)

ThePrimeagen:
if you look at `harpoon/init.lua` and search for this line `self._extensions:emit(Extensions.event_names.LIST_CREATED, list)` you can do the same thing for hooking into the buf / window created.

ThePrimeagen:
I am planning on doing another harpoon season coming up. 

I just feel pulled in every direction, it's very difficult, but I'm very excited 

I will make sure this gets somewhat resolved

tris203:
@ThePrimeagen I can reopen the PR if you think it is still useful?

It was your other direction comment made me think it was no longer relevant

pockata:
This is already possible with the events system. I've created a [small plugin](https://github.com/pockata/harpoon-highlight-current-file) that adds this functionality. The actual code is so small that you can copy it to your config and save yourself a dependency.

PR #399, opened 2023-12-09
Fix UI:select_menu_item

name-snrl:
#381

### DIFF
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -166,8 +166,9 @@ function HarpoonUI:select_menu_item(options)
         options
     )
 
-    self.active_list:select(idx, options)
+    list = self.active_list
     self:close_menu()
+    list:select(idx, options)
 end
 
 function HarpoonUI:save()


name-snrl:
@willothy @ThePrimeagen what do you think, guys?

lua/harpoon/buffer.lua
```diff
@@ -73,6 +73,20 @@ function M.setup_autocmds_and_keymaps(bufnr)
         "<Cmd>lua require('harpoon.buffer').run_select_command()<CR>",
         {}
     )
+    vim.api.nvim_buf_set_keymap(
+        bufnr,
+        "n",
+        "<C-x>",
```
willothy, line 79: Any reason why this is `<C-x>` and not `<C-s>`? I feel like it should be consistent with default split binds as long as it's hardcoded.

willothy (commented):
Looks good to me other than the keymap question, I'll clone it and test it out

lua/harpoon/buffer.lua
name-snrl: > why this is `<C-x>`

the only reason is https://github.com/ThePrimeagen/harpoon/blob/867e212ac153e793f95b316d1731f3ca1894625e/lua/harpoon/init.lua#L41

:D

willothy: Ah I see! Nevermind then haha

lua/harpoon/buffer.lua
```diff
@@ -73,6 +73,20 @@ function M.setup_autocmds_and_keymaps(bufnr)
...
+        "<Cmd>lua require('harpoon.buffer').run_select_command({ split = true })<CR>",
+        {}
+    )
+    vim.api.nvim_buf_set_keymap(
+        bufnr,
+        "n",
+        "<C-v>",
+        "<Cmd>lua require('harpoon.buffer').run_select_command({ vsplit = true })<CR>",
+        {}
+    )
```
gepbird, line 89: You could use the `vim.keymap.set` that accepts lua functions too
```suggestion
    vim.keymap.set(
        "n",
        "<C-v>",
        function() require("harpoon.buffer").run_select_command({ vsplit = true }) end,
        { buffer = bufnr }
    )
```

lua/harpoon/buffer.lua
```diff
@@ -73,6 +73,20 @@ function M.setup_autocmds_and_keymaps(bufnr)
         "<Cmd>lua require('harpoon.buffer').run_select_command()<CR>",
         {}
     )
+    vim.api.nvim_buf_set_keymap(
+        bufnr,
+        "n",
+        "<C-x>",
+        "<Cmd>lua require('harpoon.buffer').run_select_command({ split = true })<CR>",
+        {}
+    )
```
gepbird, line 82: Same as before (github doesn't let you comment on lines that have not changed)
```lua
    vim.keymap.set(
        "n",
        "q",
        function() require("harpoon.buffer").run_toggle_command("q") end,
        { silent = true, buffer = bufnr }
    )
    vim.keymap.set(
        "n",
        "<ESC>",
        function() require("harpoon.buffer").run_toggle_command("Esc") end,
        { silent = true, buffer = bufnr }
    )
    vim.keymap.set(
        "n",
        "<CR>",
        function() require("harpoon.buffer").run_select_command() end,
        { buffer = bufnr }
    )
    vim.keymap.set(
        "n",
        "<C-x>",
        function() require("harpoon.buffer").run_select_command({ split = true }) end,
        { buffer = bufnr }
    )
```

lua/harpoon/buffer.lua
gepbird, line 1: After those changes you could extract harpoon.buffer import
```lua
local Buffer = require("harpoon.buffer")
```

Or even the functions too, but I'm not sure if it's nicer
```lua
local run_toggle_command = Buffer.run_toggle_command
local run_select_command = Buffer.run_select_command
```

gepbird (changes requested):

lua/harpoon/buffer.lua
name-snrl: We could also just use callback option in `vim.api.nvim_buf_set_keymap`, but oke

lua/harpoon/buffer.lua
name-snrl: I just used an internal function

ThePrimeagen:
so i have accidentally done some of these works.  i didn't realize about this.  so we are going to have some issues.

i am trying to think through the split mappings and select options, if you want to make the change and then update the pr i am glad to look through things.

lua/harpoon/buffer.lua
```diff
@@ -52,27 +52,25 @@ function M.setup_autocmds_and_keymaps(bufnr)
...
+
+    vim.keymap.set("n", "<CR>", function()
+        M.run_select_command()
+    end, { buffer = bufnr })
+
+    vim.keymap.set("n", "<C-x>", function()
+        M.run_select_command({ split = true })
+    end, { buffer = bufnr })
+
+    vim.keymap.set("n", "<C-v>", function()
```
ThePrimeagen, line 71: so i don't want these as options within.  i think we need to figure out a way for the user to provide these through some hook.

as of now i do have a new event that i have added to listen for when the buffer is created.  I think I want to drive more and more of that for behavior, not in the code

lua/harpoon/ui.lua
```diff
@@ -145,8 +145,9 @@ function HarpoonUI:select_menu_item(options)
         options
     )
 
-    self.active_list:select(idx, options)
+    list = self.active_list
     self:close_menu()
+    list:select(idx, options)
```
ThePrimeagen, line 150: this is great.  i want this change by itself!

lua/harpoon/buffer.lua
```diff
@@ -15,10 +15,10 @@ local function get_harpoon_menu_name()
     return HARPOON_MENU .. harpoon_menu_id
 end
 
-function M.run_select_command()
+function M.run_select_command(options)
```
ThePrimeagen, line 18: i want to consider this change more carefully and provide hooks

name-snrl:
@ThePrimeagen Now it only fixes the UI:select_menu_item behavior. I'm not sure if I'll have time to work on the mappings. So you can merge this small change right now

ThePrimeagen:
pr is out of date.  issue already resolved

name-snrl:
> issue already resolved

w8, what? Problem solved? With what? I think the problem still occurs, are you sure the behavior I was talking about here is fixed?

https://github.com/ThePrimeagen/harpoon/pull/392#issuecomment-1848697502

willothy:
^ not resolved, split mappings still do not work without this fix. The menu needs to be closed before the item is selected.

PR #400, opened 2023-12-10
Feat: add replace function

th-herve:
Add a replace function that either:
- swap two existing items in the list
- replace an existing item with a new one

## example

With these key binds:
```lua
vim.keymap.set("n", "<C-j>", function() harpoon:list():select(1) end)
vim.keymap.set("n", "<C-k>", function() harpoon:list():select(2) end) 
vim.keymap.set("n", "<C-l>", function() harpoon:list():select(3) end) 

vim.keymap.set("n", "<M-j>", function() harpoon:list():replace(1) end)
vim.keymap.set("n", "<M-k>", function() harpoon:list():replace(2) end)
vim.keymap.set("n", "<M-l>", function() harpoon:list():replace(3) end)
```
And this harpoon list:
```lua
1 file_1
2 file_2
```

Pressing M-k in file_1 results in:
```lua
1 file_2
2 file_1
```

Pressing M-k in file_3 results in:
```lua
1 file_2
2 file_3
```
If there is no file harpooned at the selected index, it appends the file,
pressing M-l in file_1:
```lua
1 file_2
2 file_3
3 file_1
```


ThePrimeagen:
i am unlikely have this change.  i don't think i want this level of functionality in harpoon.  let me stew on it for a few days.

there is something about making things so fully featured that feels wrong or too complicated

daltondiaz (approved):

daltondiaz:
I liked this idea because most of the time I'm working in 3~4, one for code a feat, other test, and sometimes third and four there are files to use this feat. 

When I change the implementation I have to go in UI and remove the files, with replace this can help that you don't need to delete or move the file on the UI.

Another thing that I noted in the test is the can use replace instead of select, but this is more personal.
```lua
vim.keymap.set("n", "<leader>1", function() harpoon:list():select(1) end)
vim.keymap.set("n", "<leader>2", function() harpoon:list():select(2) end)
vim.keymap.set("n", "<leader>3", function() harpoon:list():select(3) end)
vim.keymap.set("n", "<leader>4", function() harpoon:list():select(4) end)

vim.keymap.set("n", "<leader>5", function() harpoon:list():replace(1) end)
vim.keymap.set("n", "<leader>6", function() harpoon:list():select(2) end)
vim.keymap.set("n", "<leader>7", function() harpoon:list():select(3) end)
vim.keymap.set("n", "<leader>8", function() harpoon:list():select(4) end)
```
I can use 

```
vim.keymap.set("n", "<leader>1", function() harpoon:list():replace(1) end)
vim.keymap.set("n", "<leader>2", function() harpoon:list():select(2) end)
vim.keymap.set("n", "<leader>3", function() harpoon:list():select(3) end)
vim.keymap.set("n", "<leader>4", function() harpoon:list():select(4) end)
```
Will work without select.

I want to hear more of the thoughts that Prime have about, maybe I'm not seeing something important.

ThePrimeagen:
its just hard to use :) and requires a lot of remaps, but i can be convinced on this one

dalton if you say this is a great one, maybe i just need a bit more time and could come around to using this myself.  i just think its a lot of over head :)

ThePrimeagen:
so when we add an item, we emit an event.

you need to think of events to emit here, which probably will be add, remove, and move

th-herve:
Yeah I get that it might be over complicated. The main idea is to not have to go in the ui to remove a file i'm done working with, and directly add a new one in it's place.

I added events:
- ADD and REMOVE in case a file is replaced with a new one
- REORDER in case two files are swapped

ThePrimeagen:
ok i have one problem with this

if you replace at position 3 and there are no items in the list then we have something broken.  The list's length will be 0, but there is 1 item in it (position 3).

You need a way to prevent this from happening

th-herve:
If you attempt to replace a position that is empty, the `append` function is called instead to ensure that there is no empty position in the list. 

So in this case, the item would be put at position 1 and counted in the length.

PR #404, opened 2023-12-11
fix: restore acwrite

tris203:
add definition of acwrite, this allows modifying and :w'ing the harpoon buffer

additionally, gets rid of a deprecated API call
https://neovim.io/doc/user/deprecated.html

### DIFF
--- a/lua/harpoon/buffer.lua
+++ b/lua/harpoon/buffer.lua
@@ -45,7 +45,8 @@ function M.setup_autocmds_and_keymaps(bufnr)
         vim.api.nvim_buf_set_name(bufnr, get_harpoon_menu_name())
     end
 
-    vim.api.nvim_buf_set_option(bufnr, "filetype", "harpoon")
+    vim.api.nvim_set_option_value("filetype", "harpoon", { buf = bufnr })
+    vim.api.nvim_set_option_value("buftype", "acwrite", { buf = bufnr })
     vim.keymap.set("n", "q", function()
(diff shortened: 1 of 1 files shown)


ThePrimeagen:
Could you resolve the conflict

willothy:
oops my bad, didn't think that my getting rid of the rest of the `*_set_option` calls would cause a conflict here

ThePrimeagen:
i still don't know what acwrite is and there is no :h acwrite OR i cannot telescope the help tags to find anything with acwrite in it :(

ThePrimeagen:
sorry, i need this change asap, so... i am going to just make the change myself sowwy :(

ThePrimeagen:
it wasn't taht i didn't want to take yours, i just couldn't fix a bug without it

i am adding in unit tests now to prevent this type of issue from happening again

willothy:
> i still don't know what acwrite is and there is no :h acwrite OR i cannot telescope the help tags to find anything with acwrite in it :(

`:h buftype`

```vim
acwrite	buffer will always be written with |BufWriteCmd|s

"acwrite" implies that the buffer name is not related to a file, like
"nofile", but it will be written.  Thus, in contrast to "nofile" and
"nowrite", ":w" does work and a modified buffer can't be abandoned
without saving.  For writing there must be matching |BufWriteCmd|,
|FileWriteCmd| or |FileAppendCmd| autocommands.
```

PR #406, opened 2023-12-11
feat: add ui border and title pos. settings

willothy:
Adds some ui options that can be overridden by the `toggle_opts` passed to `toggle_quick_menu`. I think being able to set defaults is nice so settings don't need to be set in keymaps.

Idea that I didn't add here but could be cool: as of nvim 0.10 floating windows can have a `footer` - we could use this to show the name of the current list for nvim nightly users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,8 +172,9 @@ Settings can alter the experience of harpoon
 ---@class HarpoonSettings
 ---@field save_on_toggle boolean defaults to false
 ---@field sync_on_ui_close boolean defaults to false
+---@field border string | string[8] defaults to "single"
+---@field title_pos "center" | "left" | "right" defaults to "center"
 ---@field key (fun(): string)
-
 ```
 
 **Descriptions**
@@ -186,6 +187,8 @@ Settings can alter the experience of harpoon
 settings = {
     save_on_toggle = false,
     sync_on_ui_close = false,
+    border = "single",
+    title_pos = "center"
     key = function()
         return vim.loop.cwd()
     end,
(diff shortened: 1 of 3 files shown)


ThePrimeagen:
@willothy i like the idea of footers

ThePrimeagen:
i am going to take care of this because i had a very different idea of how to go about this that i liked better

ThePrimeagen:
ok @willothy do you see the change?

i wanted to intentionally make this change so that you could see where i am going.  Any specific function options i want to keep as close to the function as possible.

I am going to revamp the all the settings.  Ideally there are as few settings as possible and as many events as possible with the ability to change anything needed so that people could style / alter behavior where needed.  driving behavior through config sucks :)

ThePrimeagen:
i am still thinking through how to add a more composable way for adding functionality.

more than just callbacks on things that happened, but callbacks _when_ things are about to happen and giving a "plugin" opportunity to make any changes

willothy:
> i am going to take care of this because i had a very different idea of how to go about this that i liked better

Totally! I had a feeling you'd want to implement this, no worries at all. Just added it in for my own temporary use and figured I'd send it in.

> Ideally there are as few settings as possible and as many events as possible with the ability to change anything needed so that people could style / alter behavior where needed. driving behavior through config sucks :)

This sounds great. I think the toggle opts work perfectly for the UI changes though (i.e. no events needed imo?), since the major style elements like border should be set before the window is opened to avoid flicker.

willothy:
I think Wezterm's event system is a great example to take inspiration from - it does similar things with `wezterm.on(event, handler)`, expecting handlers to have a particular side affect such as setting a tab title. 

You could do something like that with the window appearance

PR #408, opened 2023-12-11
feat: handler removal

willothy:
Restructures handlers a bit by using a map-like table instead of a list-like. The keys are the handlers, so they can easily be removed as needed (and are automatically deduplicated).

<strike> This allowed me to implement live-updating for the menu, so it will automatically update its contents if modified via Lua while it's open.</strike>

Also removed a zero from that `math.random` because it was sometimes causing an error to be thrown when naming the buffer.

If you like the handlers change but don't like the live-updating change, I'm happy to remove that.

I think it's important for handlers to be able to be removed / unregistered.

Edit: the live-update causes some weirdness so I am going to remove that and just leave this at handler removal for now.

### DIFF
--- a/lua/harpoon/listeners.lua
+++ b/lua/harpoon/listeners.lua
@@ -3,8 +3,8 @@
 --- TODO: Rename this... its an odd name "listeners"
 
 ---@class HarpoonListeners
----@field listeners (HarpoonListener)[]
----@field listenersByType (table<string, HarpoonListener>)[]
+---@field listeners table<HarpoonListener, HarpoonListener>
+---@field listenersByType table<string, table<HarpoonListener, HarpoonListener>>
 local HarpoonListeners = {}
 
 HarpoonListeners.__index = HarpoonListeners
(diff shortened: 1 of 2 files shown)


lua/harpoon/buffer.lua
```diff
@@ -8,7 +8,7 @@ local HARPOON_MENU = "__harpoon-menu__"
 -- simple reason here is that if we are deving harpoon, we will create several
 -- ui objects, each with their own buffer, which will cause the name to be
 -- duplicated and then we will get a vim error on nvim_buf_set_name
-local harpoon_menu_id = math.random(1000000)
+local harpoon_menu_id = math.random(100000)
```
daltondiaz, line 11: For experience, I'm thinking you can revert only this part of code to no cause unnecessary conflicts with your other PR https://github.com/ThePrimeagen/harpoon/pull/414/files

willothy: Yep thanks for reminding me! Forgot to do this when I made the other PR :)

willothy:
No longer applicable since listeners was changed to extensions.

ThePrimeagen:
got me

willothy:
Oops also the `update_contents` change wasn't meant to be in this PR, that was leftover from testing live-update :/

PR #410, opened 2023-12-11
feat: support prepopulating empty lists

willothy:
Suggested by @tris203

Adds support for lists which prepopulate their values when empty. This method on the list config is only run the first time the list is read from, if the list is empty. Not implemented by the default list.

This is important for lists of terminals and other things that are convenient to have pre-marked.

Another use case is pre-configuring projects by language. For example, auto-marking `src/main.rs` and `Cargo.toml` in a Rust project.

Example, prepopulate the default list with entries from `mini.visits` frecency index:

```lua
harpoon:setup({
  settings = {
    -- ...
  },
  default = {
    prepopulate = function()
      local cwd = vim.uv.cwd()
      local limit = 3
      return vim
        .iter(require("mini.visits").list_paths())
        :enumerate()
        :filter(function(i) -- because there's no `take(n)` in vim.iter
          return i <= limit
        end)
        :map(function(_, path)
          local p = Path:new(path):make_relative(cwd)
          return {
            value = p,
            context = {
              row = 1,
              col = 1,
            },
          }
        end)
        :totable()
    end,
  },
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ There is quite a bit of behavior you can configure via `harpoon:setup()`
 ---@field select? (fun(list_item?: HarpoonListItem, list: HarpoonList, options: any?): nil)
 ---@field equals? (fun(list_line_a: HarpoonListItem, list_line_b: HarpoonListItem): boolean)
 ---@field add? fun(item: any?): HarpoonListItem
+---@field prepopulate? fun(): HarpoonListItem[]
 ---@field BufLeave? fun(evt: any, list: HarpoonList): nil
 ---@field VimLeavePre? fun(evt: any, list: HarpoonList): nil
 ---@field get_root_dir? fun(): string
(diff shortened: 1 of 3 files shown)


tris203:
I have tested it too, with this config

```lua
harpoon:setup({
    tmux = {
        prepopulate = function()
            print("running prepopulate");
            local sessions = io.popen("tmux list-sessions"):read("*a")
            print(sessions)
            local lines = {}
            for s in sessions:gmatch("[^\r\n]+") do
                table.insert(lines, { value = s, context = { row = 1, col = 1} } )
            end
            return lines
        end,
        select = function(list_item, list, option)
            local sessionName = string.match(list_item.value, "([^:]+)")
            os.execute("tmux switch-client -t " .. sessionName)
        end
    }
})

```

my only comment, is that for a list like this tmux one, it would be fully managed, so I would probably want to always over write the list with whats there.

maybe we need a flag for that too

ThePrimeagen:
I like this idea, but I may want a different approach.

I would rather add a hook to allow users to manipulate a list that's just created as opposed to putting in that logic in harpoon

willothy:
> I like this idea, but I may want a different approach.
> 
> I would rather add a hook to allow users to manipulate a list that's just created as opposed to putting in that logic in harpoon

The example using the default menu is just an example, there's no default implementation of `prepopulate`, so this would just be added to user configs.

But I may be misunderstanding, do you mean an event that is dispatched instead and passes the list to the handler? I could definitely do that :)

It may be tricker to do because there isn't currently a way to listen for events only for a single list though.

ThePrimeagen:
This isn't the way i want to solve this particular problem.  Instead i am going to use hooks to solve this because i don't want more code to maintain.

Even though your feature does practically nothing and is easy to maintain, if someone else wanted something to happen when a list was created, then i would have too maintain that code too.  This becomes worse and worse.  Therefore, having hooks for you to do what you want when you want will be much better for future ThePrimeagen

willothy:
Maybe with the hooks there should be a way to register an extension for a specific list? Otherwise every single handler will be called for a list's `LIST_CREATE` event and each will need to check the lists's name... imo it should be the caller the determines the receiver of the event, not the callee.

PR #413, opened 2023-12-12
fix: check list.config.encode, default save_on_toggle to false

willothy:
Fixes lists that don't encode by checking if `list.config.encode` is nil instead of checking `list.encode` (which will never be nil).

Also actually defaults `save_on_toggle` to true, because it says it defaults to true in several places (it also says false in some places? idk) and harpoon basically doesn't work without it unless `sync_on_ui_close` is enabled.

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -14,7 +14,7 @@ M.DEFAULT_LIST = DEFAULT_LIST
 
 ---@class HarpoonPartialConfigItem
 ---@field select_with_nil? boolean defaults to false
----@field encode? (fun(list_item: HarpoonListItem): string)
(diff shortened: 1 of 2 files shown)


ThePrimeagen:
Can we please swap save on toggle to false? 

It's really weird behavior, I supported for others

willothy:
> Can we please swap save on toggle to false?
> 
> It's really weird behavior, I supported for others

Gotcha, sounds good. Wasn't sure because it said true in some places and false in others :)

willothy:
Ah I've just checked and you're totally correct that this shouldn't be false by default. The issues I've been having are related to a QuitPre autocmd I have that ends up closing nvim before VimLeavePre runs.

ThePrimeagen:
I HATE CI

PR #414, opened 2023-12-12
fix: remove a zero from the random buf name

willothy:
This can cause an error when opening the UI. Removing a single zero fixes it. This fix is included in #408 but figured I'd add it here separately so it can be merged quickly.

sorry for the PR spam lol


daltondiaz:
How do you get this error? Do you have the steps to reproduce?

willothy:
> How do you get this error? Do you have the steps to reproduce?

Huh, I've just tried again and now I can't repro it. Strange. Just gonna close this now lol. I did update nvim during the time since I posted this here though, so maybe just a weird nightly issue that was fixed. Who knows.

Issue #402, opened 2023-12-10
Popup is under going some changes


ThePrimeagen:
Highlight groups and popup window has been reworked.

we are going to drop the plenary deps and just use the nvim_open_win function.  This also means that i need to expose some border options.

as for highlight groups, there is a new event that emits out window and buf.  but i am unsure what else is needed for highlight groups

ThePrimeagen:
Yes, I would like to drive as much as possible settings through the function itself, not some distant settings

Arisa-Snowbell:
Yeah, the window looks completely different, the most annoying thing is background color is wrong because it doesn't use the highlight rn 
![image](https://github.com/ThePrimeagen/harpoon/assets/22961592/65910d8b-8eed-455b-83eb-4bbc0944d6da)

patricio-ferraggi-deel:
yeah, reverted a few commits back because I like my highlight groups 😆

PR #419, opened 2023-12-14
feat(ui): accept title_pos option if specified

ktsujister:
This is related to #402.

This let's you revert back some of the UI looks.

```lua
local harpoon = require('harpoon')
harpoon.ui:toggle_quick_menu(harpoon:list()), { border = "rounded", title_pos = "center" })
```


ThePrimeagen:
good great goodness, thank you

PR #420, opened 2023-12-14
Popup window bg_highlight

gataky:
Allows the user to pass in bg_highlight when calling to open the harpoon window. Defaults to Normal bg highlight color.


gataky:
Closing as I just saw a comment saying all highlighting will be removed and should be applied in a different way.

Issue #417, opened 2023-12-13
Keep viewport when switching files


lars-vc:
**What issue are you having that you need harpoon to solve?**
I would like to keep the viewport after switching files, currently the line is always centered on the screen.
Keeping the viewport would only be necessary in the same session (after nvim reboot this is not required)

**Why doesn't the current config help?**
No options are provided for this

**What proposed api changes are you suggesting?**
Vim already has some options to keep the viewport with winsaveview() and winrestview(view), these could be used to save the viewport.

ThePrimeagen:
you want the same position the viewport was in before you left the file?  is that what you are asking?

lpanebr:
I also would like that.

ThePrimeagen:
Okay, it seems like an okay feature. I don't really use it nor notices, so if someone would like to take a swing at saving the additional information required in the list items, context and using that on select I'd be just fine merging it in

lars-vc:
Okay, I tried some stuff but got a lil confused, so here 
https://github.com/ThePrimeagen/harpoon/blob/9031087ff1b18d0a34bd664719ec66cc8be1efd8/lua/harpoon/config.lua#L194 you save where you are at in the file when leaving a buffer correct? But when testing, it seems this code is never triggered (the logger line right before doesn't show up).

lars-vc:
This wiki is related to the problem viewport problem @lpanebr . Might just use this for consistency everywhere
https://vim.fandom.com/wiki/Avoid_scrolling_when_switch_buffers

lpanebr:
@lars-vc Thank you! This is just AWESOME and works like magic! ❤🎉

> This wiki is related to the problem viewport problem @lpanebr . Might just use this for consistency everywhere https://vim.fandom.com/wiki/Avoid_scrolling_when_switch_buffers

ThePrimeagen:
I am back, I'll review this today and merge/give feedback

PR #421, opened 2023-12-14
feat(ui): add config option to keep window when item selected

theKnightsOfRohan:
Addresses issue #417


ThePrimeagen:
i don't want this change.  Its not that i don't think it was done well, its just not what i want harpoon to do

use `prev` or `next` if you wish to peruse your list

PR #423, opened 2023-12-15
feat: show name of current list in window footer

willothy:
For users of nvim 0.10, adds the current list's name to the window's footer.


ThePrimeagen:
do we have to have specific checks?  i don't want to have vim version specific code

willothy:
> do we have to have specific checks?

Yes, because including the 'footer' / 'footer_pos' keys will throw an error before 0.10.

> i don't want to have vim version specific code

No worries, feel free to close :) 

I can just use this as an extension in my own config instead.

willothy:
Closed - there is no way to use the window footer without checking nvim version or breaking all non-nightly users.

PR #425, opened 2023-12-15
feat: add `LIST_READ` event

willothy:
Tiny change, but I think important as it would allow for some lists to be fully automated. Unsure if this should also be emitted after the list is created.

Event runs when a list is read but not for the first time (i.e. early return before `LIST_CREATE`)

### DIFF
--- a/lua/harpoon/extensions/init.lua
+++ b/lua/harpoon/extensions/init.lua
@@ -1,5 +1,3 @@
---- TODO: Rename this... its an odd name "listeners"
-
 ---@class HarpoonExtensions
 ---@field listeners HarpoonExtension[]
 local HarpoonExtensions = {}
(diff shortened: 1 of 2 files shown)


daltondiaz:
@willothy I only have a doubt, this isn't an even a review, but what do you mean when say 

>  I think important as it would allow for some lists to be fully automated

I'm new in neovim and lua env.

Thanks

tris203:
> @willothy I only have a doubt, this isn't an even a review, but what do you mean when say 
> 
> >  I think important as it would allow for some lists to be fully automated
> 
> I'm new in neovim and lua env.
> 
> Thanks

Imagine a list that isn't files. But instead is tmux sessions, or current running terminals/processes

A list you wouldn't manually edit or configure but instead would be read and provided for you to interact with @daltondiaz

ThePrimeagen:
i honestly don't understand what you are going for here, but i trust that this is useful

willothy:
> i honestly don't understand what you are going for here, but i trust that this is useful

Thank you! This is similar to what I had with `prepopulate` in #410, but without the list-specific behavior.

Would be happy to provide config examples in a later PR or in github wiki if you open that up.

PR #426, opened 2023-12-15
feat: allow user to define window height

carlesba:
I couldn't find a way to define window's height. I noticed the height was hardcoded so I mimic the config for width and apply it for height.

There's probably a better way to define height but I couldn't find that in the code.

As far as I understand, I could increase the height now using:

```lua
harpoon.ui:toggle_quick_menu(
          harpoon:list(),
          {
            ui_fallback_height = 20
          }
        )
```

### DIFF
--- a/lua/harpoon/ui.lua
+++ b/lua/harpoon/ui.lua
@@ -6,12 +6,14 @@ local Extensions = require("harpoon.extensions")
 ---@field border? any this value is directly passed to nvim_open_win
 ---@field title_pos? any this value is directly passed to nvim_open_win
 ---@field ui_fallback_width? number
+---@field ui_fallback_height? number
 ---@field ui_width_ratio? number
 
 ---@return HarpoonToggleOptions
 local function toggle_config(config)
     return vim.tbl_extend("force", {
         ui_fallback_width = 69,
+        ui_fallback_height = 8,
         ui_width_ratio = 0.62569,
     }, config or {})
 end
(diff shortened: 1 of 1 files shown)


daltondiaz:
Nice move, I think is missing in `config.lua` the property `ui_fallback_height`, so we can configure like that:

```lua
harpoon:setup({
    settings = {
        save_on_toggle = true,
        ui_fallback_height = 20
    }
})
```

carlesba:
I agree with you but those changes require a deeper refactor because right now the ui doesn't use setup settings at all.

In that case I would consider renaming that variable to `ui_height` (and `ui_width`). 

Then the priority to read the values would be
1. toggle_quick_menu
2. settings
3. fallback


wdyt?

maxrzaw:
I have been playing around with some more UI options and I came to the same realization about the config. I don't have a problem with how it is now, but either moving it to the settings or at least adding a section to the documentation for them would be nice so users don't need to read the code to find the UI settings.

As far as the naming of the UI option for the height. I think `ui_fallback_height` is a bit misleading. The existing `ui_fallback_width` is a fallback for when no window is found to calculate the height. I think `ui_height` would be more appropriate.

carlesba:
I agree that the UI config should be provided in the setup. I did this PR as a quick patch to allow setting a height because the current height is too small for my normal usage (coming from v1).

However, implementing UI settings in the setup requires a bigger implementation and I'm not sure @ThePrimeagen would approve. According to the last live-stream I watched, this issue doesn't seem a priority. 

I'll close this PR looking forward for a better implementation.

PR #432, opened 2023-12-18
Auto generate docs

### LINKED ISSUE #401: TJ HELP, DOCS IMMEDIATELY
**What issue are you having that you need harpoon to solve?**
Get tj to make ci to generate docs

**Why doesn't the current config help?**
because it doesn't make docs

**What proposed api changes are you suggesting?**
make tj make docs

dozken:
Generates vimdoc from the README.md file

Closes #401

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,25 @@
+name: Docs
+
+on: [push, pull_request]
+
+jobs:
+    docs:
+        name: Auto generate docs
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+
+            - uses: kdheepak/panvimdoc@main
+              with:
+                vimdoc: harpoon
+                pandoc: README.md
+                description: Best plugin ever created
+                demojify: true
+                treesitter: true
+                shiftheadinglevelby: -1 
+
+            - uses: stefanzweifel/git-auto-commit-action@v4
+              with:
+                commit_message: "Auto generate docs"
+                branch: ${{ github.head_ref }}
+                file_pattern: 'doc/*.txt'
--- a/doc/harpoon.txt
+++ b/doc/harpoon.txt
@@ -1,0 +1,281 @@
+*harpoon.txt*                                         Best plugin ever created
+
+==============================================================================
+Table of Contents                                  *harpoon-table-of-contents*
+
+1. ⇁ TOC                                                   |harpoon-⇁-toc|
+2. ⇁ The Problems                                 |harpoon-⇁-the-problems|
+3. ⇁ The Solutions                               |harpoon-⇁-the-solutions|
+4. ⇁ Installation                                 |harpoon-⇁-installation|
+5. ⇁ Getting Started                           |harpoon-⇁-getting-started|
+  - Quick Note                        |harpoon-⇁-getting-started-quick-note|
+  - harpoon:setup() IS REQUIRED|harpoon-⇁-getting-started-harpoon:setup()-is-required|
+  - Basic Setup                      |harpoon-⇁-getting-started-basic-setup|
+6. ⇁ API                                                   |harpoon-⇁-api|
+  - Config                                            |harpoon-⇁-api-config|
+  - Settings                                        |harpoon-⇁-api-settings|
+  - Highlight Groups                        |harpoon-⇁-api-highlight-groups|
+  - Logger                                            |harpoon-⇁-api-logger|
+7. ⇁ Contribution                                 |harpoon-⇁-contribution|
+8. ⇁ Social                                             |harpoon-⇁-social|
+9. ⇁ Note to legacy Harpoon 1 users|harpoon-⇁-note-to-legacy-harpoon-1-users|
+
+
+==============================================================================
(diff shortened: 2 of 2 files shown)


doc/harpoon.txt
```diff
@@ -0,0 +1,281 @@
+*harpoon.txt*                                         Best plugin ever created
+
+==============================================================================
+Table of Contents                                  *harpoon-table-of-contents*
+
+1. ⇁ TOC                                                   |harpoon-⇁-toc|
+2. ⇁ The Problems                                 |harpoon-⇁-the-problems|
+3. ⇁ The Solutions                               |harpoon-⇁-the-solutions|
+4. ⇁ Installation                                 |harpoon-⇁-installation|
+5. ⇁ Getting Started                           |harpoon-⇁-getting-started|
```
willothy, line 10: Do you think something can be done about these tags? I don't think including the symbols here is great...

doc/harpoon.txt
```diff
@@ -0,0 +1,281 @@
...
+- `[name] = HarpoonPartialConfigItem`any named lists config. it will be merged with `default` and override any behavior
+
+**HarpoonPartialConfigItem Definition**
+
+>
+    ---@class HarpoonPartialConfigItem
+    ---@field encode? (fun(list_item: HarpoonListItem): string)
+    ---@field decode? (fun(obj: string): any)
+    ---@field display? (fun(list_item: HarpoonListItem): string)
+    ---@field select? (fun(list_item?: HarpoonListItem, list: HarpoonList, options: any?): nil)
```
willothy, line 184: I think these should be converted to vimdoc style if possible instead of a code block, probably fine if there's no convenient way to do that though.

doc/harpoon.txt
dozken: an easy solution would be to remove them from README.md

.github/workflows/docs.yml
```diff
@@ -0,0 +1,25 @@
...
+            - uses: actions/checkout@v4
+
+            - uses: kdheepak/panvimdoc@main
+              with:
+                vimdoc: harpoon
+                pandoc: README.md
+                description: Best plugin ever created
+                demojify: true
+                treesitter: true
+                shiftheadinglevelby: -1
```
dozken, line 19: @willothy this step generates the vimdoc from the README.md file.
which requires writing the markdown in a way that will be better converted to the vimdoc

From the comments above I can propose these options:
1. create another markdown file e.g. DOC.md which will contain documentation
2. update README.md to support seamless conversion to  vimdoc
3. write docs in the lua files and generate docs from there as it is done in https://github.com/nvim-telescope/telescope.nvim/blob/master/scripts/gendocs.lua

doc/harpoon.txt
dozken: answered here https://github.com/ThePrimeagen/harpoon/pull/432#discussion_r1430832152

doc/harpoon.txt
willothy: Yeah, I wonder how prime would feel about that. Otherwise, is `panvimdoc ignore ...` allowed inline, or is that only able to ignore multiple lines? It *may* be possible to use ignore to keep the symbols in the markdown file and ignore them in the vimdoc, but I'm not sure. I'll look at the panvimdoc configs for my plugins and see if I've had to do something similar.

.github/workflows/docs.yml
willothy: Yeah, I guess it really depends on what prime prefers, but imo the simplest solution would be to just update the readme a bit. Overall though this is a great PR and I think autogenerated docs will be super helpful for everyone :)

dozken: so the next course of action is to wait for @ThePrimeagen's response, right?

willothy: Yep :)

.github/workflows/docs.yml
```diff
@@ -0,0 +1,25 @@
...
+        name: Auto generate docs
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+
+            - uses: kdheepak/panvimdoc@main
+              with:
+                vimdoc: harpoon
+                pandoc: README.md
+                description: Best plugin ever created
```
davvid, line 16: The description here should probably say, "Getting you where you want with the fewest keystrokes"

.github/workflows/docs.yml
```diff
@@ -0,0 +1,25 @@
...
+                demojify: true
+                treesitter: true
+                shiftheadinglevelby: -1 
+
+            - uses: stefanzweifel/git-auto-commit-action@v4
+              with:
+                commit_message: "Auto generate docs"
+                branch: ${{ github.head_ref }}
+                file_pattern: 'doc/*.txt'
```
davvid, line 25: I believe you'll need to add:
```yaml
    permissions:
      contents: write
```
under the `docs` section (at the same level as `runs-on:`) in order to use the `git-auto-commit-action`.

davvid (commented):
Also, you may want to consider amending the commits to add a `Closes: #401` message at the bottom so that the commit is marked as closing #401.

PR #442, opened 2023-12-23
Added telescope section in README

TheSantacloud:
I'm using telescope for all my navigation purposes, and this is how I did it for myself. I thought other people could benefit from it as well

I didn't make it an extension because this was enough for me.


psaikido:
Very nice!

Msouza91:
@dormunis How do you clear your harpoon list using this setup? It does look good, but when you need to switch your files might be a bit clunkier I think

PR #449, opened 2023-12-26
Remove Lua

wartab:


wartab:
Next step is porting this to Rust.

PR #451, opened 2023-12-27
feat: navigate_with_numbers

otavioschwanck:


ThePrimeagen:
this is cool, i don't think i'll use this, but i'll accept it as a "official" extension

PR #452, opened 2023-12-27
feat: Add toggle

otavioschwanck:


ThePrimeagen:
could you explain why this is good?

this seems a bit confusing why you would ever want this (you have to remember the state in your head or check the state, i would expect you, the user, to add (does nothing on dupe) or remove (does nothing on missing))

otavioschwanck:
> could you explain why this is good?
> 
> this seems a bit confusing why you would ever want this (you have to remember the state in your head or check the state, i would expect you, the user, to add (does nothing on dupe) or remove (does nothing on missing))

i added to my statusline an text to see if the current file is harpooned or not, so, i can tell when i am pinning or not.

ThePrimeagen:
Please add a test and I can make this happen

Issue #454, opened 2023-12-27
Harpoon2: Add events around UI_CREATE


pockata:
**What issue are you having that you need harpoon to solve?**
I'm trying to re-implement the functionality of #199 as a harpoon2 plugin but two more events are needed for it, e.g. `BEFORE_CREATE_UI` and `AFTER_CREATE_UI`.

**Why doesn't the current config help?**
The event `UI_CREATED` is called too early to move the cursor to the line containing the current file because the text is not in the buffer yet. And is also called too late because the focus is already set to the floating buffer so we lose the context of the buffer from which we opened the quick menu (no way to get the current file).

Currently, for it to work, a user has to change the keybinding for the `toggle_quick_menu` and surround it with these functions.
```lua
vim.keymap.set("n", "<leader>tc", function()
	require("harpoon-highlight-current-file").mark_current_file()
	harpoon.ui:toggle_quick_menu(harpoon:list())
	require("harpoon-highlight-current-file").highlight_current_file()
end)
```
This is not ideal as you might open the quick menu with a different method (perhaps another plugin or something) and not activate this functionality.

**What proposed api changes are you suggesting?**
Call `AFTER_CREATE_UI` in `toggle_quick_menu` after `nvim_buf_set_lines` and `BEFORE_CREATE_UI` in `_create_window` before `nvim_open_win`. 

I can prepare a PR if this is something of interest.

ThePrimeagen:
How about we just have UI create be called after everything is created.

Instead of a whole bunch of events, let's just try to keep it simple.

Since it sounds like you've done the research, if you just want PR moving that function call. I'll be glad to merge it

pockata:
Just to clarify, moving `UI_CREATE` is only half the solution. I would still need `BEFORE_UI_CREATE` (I'm open to name suggestions) to be called before the context is moved from the file buffer to the harpoon quick menu buffer. Otherwise I wouldn't have access to the current file.

I've submitted the PR in #457

ThePrimeagen:
Can we just include that information in the event?

pockata:
This also works. I've updated the PR.

PR #457, opened 2023-12-28
Adjust UI_CREATE events

pockata:
As discussed in #454, I've moved `UI_CREATE` to fire after the contents of the buffer have been added and created a `BEFORE_OPEN` event.


pockata:
I've updated the PR to send the current file as a parameter in `UI_CREATE` and removed the `BEFORE_OPEN` event.

PR #458, opened 2023-12-28
[fix] Application errors on first use

aaronhallaert:
On first use of the application, harpoon errors on `c_config.projects` being nil

* Add nil check


ThePrimeagen:
Harpoon 1 is frozen. Please use harpoon 2

That branch is an active development

PR #422, opened 2023-12-15
Add lazy.nvim installation instructions to README

Drahlous:
Just adding lazy.nvim instructions to the README to make it easier for folks to copy/paste into their config.

PR #460, opened 2023-12-29
docs: document lazy.nvim instalation

piraz:
It seems like the documentation is kind of repetitive but it took a little time to get the right configuration while I was migrating from packer to lazy.nvim.

On this case the `requires` will be replaced by `dependencies`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,14 @@ use {
     requires = { {"nvim-lua/plenary.nvim"} }
 }
 ```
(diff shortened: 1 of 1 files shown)


itsfrank:
Similar PR was closed a week ago, see https://github.com/ThePrimeagen/harpoon/pull/422#issuecomment-1869715778

ThePrimeagen:
i am going to do this and delete the packer one at some point.

the documentation is going to get a complete redo soon

PR #461, opened 2023-12-29
Added back the ui_nav_wrap feature

krsntn:
Can we bring back the option to turn wrapping on/off when going through harpoon lists in [#387](https://github.com/ThePrimeagen/harpoon/pull/387)? It got removed in https://github.com/ThePrimeagen/harpoon/commit/fa73cd0c33f5a4d8b0084c70da1e48169f0053f8 and I'm not sure why.

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -63,6 +63,8 @@ function M.get_default_config()
         },
 
         default = {
+            --- ui_nav_wrap allows the ability to enable(true) or disable(false) wrapping on prev and next list calls.
+            ui_nav_wrap = true,
 
             --- select_with_nill allows for a list to call select even if the provided item is nil
             select_with_nil = false,


ThePrimeagen:
@krsntn it has been moved to the toggle method.  i am trying to drive system options as small as possible (they are annoying to keep track of)

https://github.com/ThePrimeagen/harpoon/blob/harpoon2/lua/harpoon/list.lua#L188

kimabrandt-flx:
> ...
> https://github.com/ThePrimeagen/harpoon/blob/harpoon2/lua/harpoon/list.lua#L188

Here is the hardlink:
https://github.com/ThePrimeagen/harpoon/blob/9e869aae22ad95f350b6b079dedf44ead1e448e8/lua/harpoon/list.lua#L188-L201

Issue #466, opened 2023-12-31
Include more context when displaying


maxrzaw:
**What issue are you having that you need harpoon to solve?**  
I would like to be able to show the marks relative to my current file.

**Why doesn't the current config help?**  
When display is called, the current buffer is the harpoon UI buffer, so I have no way to know where it was called from.

**What proposed api changes are you suggesting?**  
Include the bufname or filename or some way to get the location of where `toggle_quick_menu()` was called from in `HarpoonList:display()` and `HarpoonPartialConfigItem.display()`
I think `HarpoonList:resolve_displayed()` might also need some tweaks.
A context that could be passed in from toggle_quick_menu and gets passed along to the display function would also help me solve my issue.

maxrzaw:
I have a working implementation, but I am not a huge fan of it because I am passing the `ui_context` into `HarpoonList:get_by_display()`, `HarpoonList:resolve_displayed()`, and `HarpoonList:display()`. I have a couple thoughts about this. `HarpoonList:get_by_display()` and `HarpoonList:resolve_displayed()` only need it because they pass it in to `HarpoonList:display()` to get the display values. This has been a bit tricky to work with and I think it would make more sense for there to be a way for the ui to identify the list items rather than the list calling display methods.

I have my WIP #467 showing some changes, but there is more to be done and I would like feedback.

ThePrimeagen:
Over the next couple weeks, I'll look into this. I want to understand the problem more deeply, as I'm a bit worried I am adding complexity for some very niche case

I also like to try to drive functionality via extensions. This allows me not to have to maintain any behavior I do not wish

PR #467, opened 2024-01-01
wip: add ui_context

maxrzaw:
I've been playing around trying to make a custom list that will show the files relative to my current context. I would like to accomplish this by passing in context to the UI when I call my keymap. This works well because I can pass in the context I need to evaluate how I want to display it when I call my keymap.

```lua
vim.keymap.set("n", "<leader>h", function()
    local path = vim.api.nvim_buf_get_name(vim.api.nvim_get_current_buf())
    -- ^ this wont work when called in HarpoonList:display() because it is the
    -- harpoon buffer which is in a different directory than the file I am editing
    harpoon.ui:toggle_quick_menu(harpoon:list("relative"), {
        context = path
    })
end)
```

This is working well for my use case, but is a WIP for a couple reasons:
1. I haven't done any unit tests
2. I think the display methods in `HarpoonList` need to be refactored into the ui so we have access to the context without having to pass it into `HarpoonList`.
3. Not sure about other use cases

This is related to issue #466

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -17,7 +17,7 @@ M.DEFAULT_LIST = DEFAULT_LIST
 ---@field select_with_nil? boolean defaults to false
 ---@field encode? (fun(list_item: HarpoonListItem): string) | boolean
 ---@field decode? (fun(obj: string): any)
----@field display? (fun(list_item: HarpoonListItem): string)
+---@field display? (fun(ui_context: any, list_item: HarpoonListItem): string)
 ---@field select? (fun(list_item?: HarpoonListItem, list: HarpoonList, options: any?): nil)
 ---@field equals? (fun(list_line_a: HarpoonListItem, list_line_b: HarpoonListItem): boolean)
 ---@field create_list_item? fun(config: HarpoonPartialConfigItem, item: any?): HarpoonListItem
@@ -80,7 +80,7 @@ function M.get_default_config()
             end,
 
             ---@param list_item HarpoonListItem
-            display = function(list_item)
+            display = function(ui_context, list_item)
                 return list_item.value
             end,
 
--- a/lua/harpoon/list.lua
+++ b/lua/harpoon/list.lua
@@ -120,8 +120,15 @@ function HarpoonList:get(index)
     return self.items[index]
 end
 
-function HarpoonList:get_by_display(name)
-    local displayed = self:display()
+--TODO: I don't think it makes sense to have these display methods in the list
+--module. I think they should be in the UI module, the list module should not
+--have anything to do with the display value. IMHO
+
+---@param ui_context any
+---@param name string
+---@return HarpoonItem|nil
+function HarpoonList:get_by_display(ui_context, name)
+    local displayed = self:display(ui_context)
     local index = index_of(displayed, name)
     if index == -1 then
         return nil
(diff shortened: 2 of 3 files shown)


maxrzaw:
Here is my config that is going with it. I haven't completely ironed out the display logic for when to show `../dir/file` vs `/complete/path/to/file`, but this is how I am using it.

```lua
-- Function to find the root directory of the project
local function find_project_root()
    ---@type string
    local current_dir = vim.uv.cwd()
    local marker_files = { ".git", "package.json", ".sln" }

    -- Check each parent directory for the existence of a marker file or directory
    while current_dir ~= "/" do
        for _, marker in ipairs(marker_files) do
            local marker_path = current_dir .. "/" .. marker
            if vim.fn.isdirectory(marker_path) == 1 or vim.fn.filereadable(marker_path) == 1 then
                return current_dir
            end
        end
        current_dir = vim.fn.resolve(current_dir .. "/..")
    end
    -- If no marker file or directory is found, return the original directory
    return vim.uv.cwd()
end

return {
    "ThePrimeagen/harpoon",
    branch = "harpoon2",
    dev = true,
    lazy = false,
    dependencies = { "nvim-lua/plenary.nvim" },
    config = function()
        local Harpoon = require("harpoon")
        local default_config = Harpoon.config.default

        Harpoon:setup({
            settings = {
                save_on_toggle = true,
                sync_on_ui_close = true,
                key = function()
                    return find_project_root()
                end,
            },
            relative = {
                select = function(list_item, list, options)
                    default_config.select(list_item, list, options)
                end,
                get_root_dir = function()
                    return find_project_root()
                end,
                equals = function(a, b)
                    return a.value == b.value
                end,
                create_list_item = function(config, name)
                    local path = vim.api.nvim_buf_get_name(vim.api.nvim_get_current_buf())
                    name = name or path
                    if string.sub(name, 1, 1) ~= "/" then
                        local dir = Path:new(path):parent().filename
                        path = dir .. "/" .. name
                    end
                    local bufnr = vim.fn.bufnr(path, false)

                    local pos = { 1, 0 }
                    if bufnr ~= -1 then
                        pos = vim.api.nvim_win_get_cursor(0)
                    end

                    return {
                        value = path,
                        context = {
                            row = pos[1],
                            col = pos[2],
                        },
                    }
                end,
                display = function(ui_context, list_item)
                    local dir = Path:new(ui_context):parent().filename
                    return Path:new(list_item.value):make_relative(dir)
                end,
                BufLeave = function(arg, list)
                    local bufnr = arg.buf
                    local bufname = vim.api.nvim_buf_get_name(bufnr)
                    local item = nil
                    for _, it in ipairs(list.items) do
                        local value = it.value
                        if value == bufname then
                            item = it
                        end
                    end

                    if item then
                        local pos = vim.api.nvim_win_get_cursor(0)

                        item.context.row = pos[1]
                        item.context.col = pos[2]
                    end
                end,
            },
        })

        -- Harpoon
        vim.keymap.set("n", "<leader>m", function()
            Harpoon:list("relative"):append()
        end)
        vim.keymap.set("n", "<leader>h", function()
            local path = vim.api.nvim_buf_get_name(vim.api.nvim_get_current_buf())
            Harpoon.ui:toggle_quick_menu(Harpoon:list("relative"), {
                border = "rounded",
                title_pos = "center",
                context = path,
            })
        end)
    end,
}
```

pockata:
The `UI_CREATE` event provides a `current_file` context field (introduced in #457). You can see it in action [here](https://github.com/pockata/harpoon-highlight-current-file/blob/223fc03c48c881340952c54bc8c63b9d27d67219/lua/harpoon-highlight-current-file/init.lua#L12).
Perhaps it can be of service?

Issue #441, opened 2023-12-22
Harpoon doesn't save cursor position after quit


romado77:
I've noticed an issue with Harpoon related to the persistence of cursor positions. When a file is harpooned, the cursor position is saved upon quitting neovim (I could see this from harpoon.json). However, when reopening the harpooned file and moving the cursor to a new position, the Harpoon doesn't preserve the updated cursor position if I quit from neovim again. So every time I open harpooned file it jumps to a position which was initially saved when file was harpooned.

My configuration:
`
local harpoon = require("harpoon")

harpoon:setup({ settings = {
	save_on_toggle = true,
} })

vim.keymap.set("n", "<leader>a", function()
	harpoon:list():append()
end, { desc = "Add current file to harpoon" })
vim.keymap.set("n", "<leader>h", function()
	harpoon.ui:toggle_quick_menu(harpoon:list())
end, { desc = "Toggle harpoon menu" })

vim.keymap.set("n", "<C-1>", function()
	harpoon:list():select(1)
end, { desc = "Select harpoon item 1" })
vim.keymap.set("n", "<C-2>", function()
	harpoon:list():select(2)
end, { desc = "Select harpoon item 2" })
vim.keymap.set("n", "<C-3>", function()
	harpoon:list():select(3)
end, { desc = "Select harpoon item 3" })
vim.keymap.set("n", "<C-4>", function()
	harpoon:list():select(4)
end, { desc = "Select harpoon item 4" })
`

romado77:
I did a bit of troubleshooting and identified a section of code in the Lua script that handles the BufLeave event. Below function `get_by_display` consistently returns -1. The argument passed to `get_by_display` is the harpoon menu name rather than the buffer name, causing the function to be unable to find a matching value in the list of harpooned files, hence never run update position.

```lua
            BufLeave = function(arg, list)
                local bufnr = arg.buf
                local bufname = vim.api.nvim_buf_get_name(bufnr)
                local item = list:get_by_display(bufname)

                if item then
                    local pos = vim.api.nvim_win_get_cursor(0)

                    Logger:log(
                        "config_default#BufLeave updating position",
                        bufnr,
                        bufname,
                        item,
                        "to position",
                        pos
                    )

                    item.context.row = pos[1]
                    item.context.col = pos[2]
                end
            end,

```

theKnightsOfRohan:
I can replicate this issue. Here is a video, in two parts because size requirements:

[...]

PR #469, opened 2024-01-02
feat: update backing value for default config

maxrzaw:
This will help with files position not being saved when you harpoon a file that is relative path from the cwd and is not a path from root and then leave the file without marking it again.

- fix unit test swap file issue
- add test for relative files
- add test for context not being saved

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -19,7 +19,7 @@ M.DEFAULT_LIST = DEFAULT_LIST
 ---@field decode? (fun(obj: string): any)
 ---@field display? (fun(list_item: HarpoonListItem): string)
 ---@field select? (fun(list_item?: HarpoonListItem, list: HarpoonList, options: any?): nil)
----@field equals? (fun(list_line_a: HarpoonListItem, list_line_b: HarpoonListItem): boolean)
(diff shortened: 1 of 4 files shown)


maxrzaw:
This addresses one of the issues mentioned in the comments of #441. Does not address the issue of lists not being saved on quit.

ThePrimeagen:
I'm going to try to look closer at this. I may merge this though. I worry it's going to cause issues. I don't like the two levels of translation.

maxrzaw:
Hold off on that. I might have found a bug in the create_list_item function.

maxrzaw:
Ok, turns out this bug already exists in the harpoon2 branch #476. This change makes it slightly different, but it still happens.

Basically the case is where I have an item in the list and I create a duplicate.
The issue I am having on this branch is that when save the UI with two items that are the same value after they are normalized, i.e ("./README.md" and "README.md"), they are both saved, which results in a duplicate in the list. On this branch they are then displayed the same the next time around which results in similar issues to #476. The problem on this branch is that two values may not be displayed the same when they are first input, but they could be the same file.

I did find another bug with the path normalization and I'll push a fix and some tests for that. Getting more familiar with `Plenary.Path`.

I'm going to think about how we could solve this issue, but at the moment, I am not sure how without making a large refactor.

PR #471, opened 2024-01-02
Harpoon 2: Improve telescope setup documentation

johnpgr:
The current documentation in `README.md` on setting Harpoon 2 with telescope is incomplete.
It does not open the buffer you selected in the picker with `<CR>`
It does not have any way to remove a buffer from the list using the picker.

The code snippet added to `README.md` in my commit is going to help users that would like to use Harpoon 2 with telescope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ In order to use [Telescope](https://github.com/nvim-telescope/telescope.nvim) as
 make sure to add `telescope` to your dependencies and paste this following snippet into your configuration.
 
 ```lua
+local finders = require("telescope.finders")
 local actions = require("telescope.actions")
 local action_state = require("telescope.actions.state")
 local harpoon = require('harpoon')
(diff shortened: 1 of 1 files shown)


nirmitjoshi:
Quick question: To remove the clutter in readme can't we just  modify and make [marks.lua](https://github.com/ThePrimeagen/harpoon/blob/harpoon2/lua/telescope/_extensions/marks.lua) requirable?

ThePrimeagen:
@nirmitjoshi mark.lua is legacy harpoon. Harpoon 2 Is the new harpoon.

If you modify that, I will not merge it :)

nirmitjoshi:
@greidinger-reis can you also add default keymaps for re-arranging files,  just to make the functionality more clear

johnpgr:
@nirmitjoshi I'm trying to get that done, but telescope doesn't provide any functions to easily re-arrange list items, so I'm having some trouble implementing my own, since I'm relatively new to telescope

johnpgr:
It looks like there is an implementation of that in `marks.lua` I'll try to copy that

johnpgr:
There you go. I added a mark rearranging feature to the PR. However this makes this setup a lot longer, it might be time to rewrite it into a harpoon2 telescope plugin.
I also left some comments in it, that might not be adequate to be in a readme file. If there's a problem with that I'll remove them

bryanvaz:
+1
Manually copied the config, and now Harpoon2+Telescope is almost as awesome as TJ's doggo

ilan-schemoul:
Hello,
@ThePrimeagen I feel like what @johnpgr offers makes sense : this should be its own plugin. We don't want people copy pasting + maintaining manually +200 lines of codes for a basic feature.

Furthermore this issue of partial telescope support, incomplete readme.md, no native support for terminals make me wonder why v1 is "deprecated". Active development in on v2 but v2 seems experimental. Also you may want to add a tag or a branch for v1 so people don't have unexpected breaking changes.

Anyway for anyone wanting to switch from v2 to v1 if you a bug saying can't derefence config of nil init.lua:270 (or something like that) you should `rm ~/.local/share/nvim/harpoon.json` (you'll lose harpoon data)

rubiin (approved):

gvolpe:
I've tried this and it almost works. The remaining items after re-opening Telescope represent an incomplete list.

For instance, if you initially have 5 items, and delete just one, then close Telescope and re-open it once again, you will find only 2 items in the list, when there should be 4 items.

Anyone else experiencing the same issue?

gvolpe:
This works great: https://github.com/ThePrimeagen/harpoon/pull/512

johnpgr:
@gvolpe Yes, I noticed this bug, I got it working by using a indexOf helper function to get the index of the mark in the list correctly. Here it is if you want, and I will include it in this PR
```lua
local function list_indexOf(list, predicate)
    for i, v in ipairs(list) do
        if predicate(v) then
            return i
        end
    end
    return -1
end
```

Use it in the `C-d` keymap
```lua
map({ 'n', 'i' }, '<C-d>', function(prompt_bufnr)
                local curr_picker = action_state.get_current_picker(prompt_bufnr)
                curr_picker:delete_selection(function(selection)
                    local mark_idx = list_indexOf(harpoon_files.items, function(v)
                        return v.value == selection[1]
                    end)
                    if mark_idx == -1 then
                        return
                    end

                    harpoon:list():removeAt(mark_idx)
                end)
            end)
```

PR #481, opened 2024-01-10
normalize buffer names when saving the buffer position

LnL7:
I'm not sure what's special about my setup, but `vim.api.nvim_buf_get_name` always returns the absolute path of a buffer. Including when no plugins are enabled. In case the bufname is already relative nothing changes so this should safely cover both cases.

This code also does not get called during VimLeavePre, as a result the position of the last buffer might not be persisted, not sure if this is intentional.
https://github.com/ThePrimeagen/harpoon/blob/harpoon2/lua/harpoon/init.lua#L148

Perhaps this also fixes the problem described in https://github.com/ThePrimeagen/harpoon/issues/441?

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -181,10 +181,12 @@ function M.get_default_config()
                 }
             end,
 
-            BufLeave = function(arg, list)
+            BufLeave = function(arg, list, config)
                 local bufnr = arg.buf
                 local bufname = vim.api.nvim_buf_get_name(bufnr)
-                local item = list:get_by_display(bufname)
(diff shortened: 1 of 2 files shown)


maxrzaw:
`vim.api.nvim_buf_get_name` always returns the full file name for the buffer. I believe this will solve some of the issues mentioned in #441, but it not being called in VimLeavePre is still an issue.

LnL7:
> `vim.api.nvim_buf_get_name` always returns the full file name for the buffer.

Strange that only a few people seem to be reporting this then. Keeping track of the position, like marks, was kind of the sales pitch for this plugin.

PR #494, opened 2024-01-18
Cross platform paths

tusqasi:
Adds platform specific path separators.

### DIFF
--- a/lua/harpoon/data.lua
+++ b/lua/harpoon/data.lua
@@ -2,7 +2,6 @@ local Path = require("plenary.path")
 
 local data_path = vim.fn.stdpath("data")
 local full_data_path = string.format("%s%s%s",data_path, Path.sep, "harpoon.json")
--- local full_data_path = string.format("%s/harpoon.json", data_path)
 
 ---@param data any
 local function write_data(data)
--- a/lua/harpoon/data.lua
+++ b/lua/harpoon/data.lua
@@ -1,7 +1,8 @@
 local Path = require("plenary.path")
 
 local data_path = vim.fn.stdpath("data")
-local full_data_path = string.format("%s/harpoon.json", data_path)
+local full_data_path = string.format("%s%s%s",data_path, Path.sep, "harpoon.json")
+-- local full_data_path = string.format("%s/harpoon.json", data_path)
 
 ---@param data any
 local function write_data(data)


tusqasi:
I have tested this on windows 10 and on my manjaro setup. 

Works like a charm✨

davvid:
Curious, is this mostly aesthetic or does this actually make a difference in behavior? I was under the impression that Windows allows `/` and `\` to be used interchangeably, which is why I ask. My naive understanding of Windows suggests that this patch is mostly aesthetic, but please correct me if I'm misunderstanding.

Here's me being an elitist Linux/UNIX snob: adding support for Windows always makes code gross and ugly. (Sorry ~ just stating facts.)

What about when Windows users have absolute paths in their harpoon list? I recall patching harpoon to handle absolute paths, and we did so by checking whether the string starts with `/`.

This PR doesn't seem to address that part, and that part is more than aesthetic since it actually affects functionality.

How gross does the code become if it were to also support absolute paths on Windows? It's probably worth seeing that fully fleshed out before committing to supporting inferior development environments IMO.

tusqasi:
> Curious, is this mostly aesthetic or does this actually make a difference in behavior? 


This change works for me.




> I was under the impression that Windows allows / and \ to be used interchangeably, which is why I ask. My naive understanding of Windows suggests that this patch is mostly aesthetic, but please correct me if I'm misunderstanding.

I too was under the impression that this should work. 





I unfortunately have to use windows on the regular for some time.

But now I use neovim nightly on wsl,So doing anything is less painful now, and no errors.



> How gross does the code become if it were to also support absolute paths on Windows? It's probably worth seeing that fully fleshed out before committing to supporting inferior development environments IMO.

IMO any code worthy of existing is ugly.




>  It's probably worth seeing that fully fleshed out before committing to supporting inferior development environments IMO.


Better to have something which works ootb for people, IMO.





I would have loved to not open the harpoon codebase when using it for the first time.

davvid:
> This change works for me.

I'm going to assume that you meant, "This change /makes it/ work for me." Thanks for confirming that it's not just aesthetic.

There are open issues about supporting absolute windows paths IIRC. It'd be great to extend this initial patch to resolve that issue so that we can truly say that harpoon supports Windows-style paths.

tusqasi:
Surely, that will be great

PR #496, opened 2024-01-19
Add harpoon list clearing to README

wilkinson4:
First time harpoon user here, and as a new user I had to do some digging to find out how to clear the harpoon list. So I figured adding it to the README would be helpful for future first timers. Thank you Primeagen for the great plugin!


wilkinson4:
Note: I didn't know what keybindings to put in there because I am a soy dev that doesn't use a dvorak layout 😅.

Issue #491, opened 2024-01-18
Problem with Harpoon + Telescope integration on Windows


umutondersu:
With the setup on the readme whenever I use the telescope harpoon menu and choose a buffer to go, instead it leads me to an empty buffer with a location "C". 

See image:
![image](https://github.com/ThePrimeagen/harpoon/assets/93402298/291d04fa-8ede-4730-a74b-517aaaae99fe)
Moreover, this problem is only on windows as in WSL with the same setup this isn't an issue.

This is my current Harpoon2 Setup with lazy.vim:
```lua
return {
    "ThePrimeagen/harpoon",
    branch = "harpoon2",
    dependencies = { "nvim-lua/plenary.nvim", 'nvim-telescope/telescope.nvim', },
    config = function()
        local harpoon = require('harpoon')
        harpoon:setup({})

        -- basic telescope configuration
        local conf = require("telescope.config").values
        local function toggle_telescope(harpoon_files)
            local file_paths = {}
            for _, item in ipairs(harpoon_files.items) do
                table.insert(file_paths, item.value)
            end

            require("telescope.pickers").new({}, {
                prompt_title = "Harpoon",
                finder = require("telescope.finders").new_table({
                    results = file_paths,
                }),
                previewer = conf.file_previewer({}),
                sorter = conf.generic_sorter({}),
            }):find()
        end

        vim.keymap.set("n", "<leader><Space>", function() toggle_telescope(harpoon:list()) end,
            { desc = "Open harpoon window" })

        vim.keymap.set("n", "<leader>x", function() harpoon:list():append() end, { desc = "Harpoon Mark" })
        vim.keymap.set("n", "<leader>X", function() harpoon:list():remove() end, { desc = "Harpoon Unmark" })
        -- vim.keymap.set("n", "<leader><Space>", function() harpoon.ui:toggle_quick_menu(harpoon:list()) end,
        --     { desc = "Open harpoon window" })

        vim.keymap.set("n", "<M-1>", function() harpoon:list():select(1) end)
        vim.keymap.set("n", "<M-2>", function() harpoon:list():select(2) end)
        vim.keymap.set("n", "<M-3>", function() harpoon:list():select(3) end)
        vim.keymap.set("n", "<M-4>", function() harpoon:list():select(4) end)
        vim.keymap.set("n", "<M-5>", function() harpoon:list():select(5) end)
        vim.keymap.set("n", "<M-6>", function() harpoon:list():select(6) end)
[...]

PR #497, opened 2024-01-19
Fix to the issue #491

umutondersu:


lua/harpoon/config.lua
```diff
@@ -164,7 +164,7 @@ function M.get_default_config()
                     )
 
                 Logger:log("config_default#create_list_item", name)
-
+                name = vim.fs.normalize(name)
```
ThePrimeagen: So how exactly does this fix it?

And if it does fix it, can we remove the normalize call above?

Let's just have one spot we call normalize.

umutondersu: I think the problem is that on windows, path normalization cannot be done correctly for the telescope menu. This solution fixes that problem with the somewhat new vim.fs.normalize() method. However only using the normalize method breaks the marking and unmarking for some reason so I'll come up with a solution where normalization is done in one place without that issue.

umutondersu:
Okay turns out name variable always returns nil on windows, so the value of name is dependent on the normalize_path function. Therefore, I wrapped the return of normalize_path with the normalize method I implemented the commit before. Now it is functioning without a problem on both windows and WSL ubuntu.

wurli:
I'm using windows 11 (I don't have a choice - it's my work machine). Unfortunately this change doesn't seem to fix the issue for me :(

ozzy1136:
> I'm using windows 11 (I don't have a choice - it's my work machine). Unfortunately this change doesn't seem to fix the issue for me :(

Try again by making the change to the normalize_path return statement. I'm on windows 11 and telescope integration now works for me.

Issue #444, opened 2023-12-24
problem with vue


momozahara:
create project using vite add multiple vue file open file 1 open file 2 then error

```
E5108: Error executing lua BufWinEnter Autocommands for "*"..function <SNR>11_Highlight_Matching_Pair, line 149: Vim(call):E801: ID already taken: 3                                                             
stack traceback:                                                                                                                                                                                                 
        [C]: in function 'bufload'                                                                                                                                                                               
        ...ut/.local/share/nvim/lazy/harpoon/lua/harpoon/config.lua:110: in function 'select'                                                                                                                    
        ...nbut/.local/share/nvim/lazy/harpoon/lua/harpoon/list.lua:182: in function 'select'                                                                                                                    
        [string ":lua"]:1: in main chunk                      
```

kimabrandt-flx:
I see the same error, when trying to open a file with associated swap-file.

When opening the first item in a list: `:lua require("harpoon"):list():select(1)`, I get the following message and error:
```
Found a swap file by the name "....swp"
          owned by: ...
          ...
      NEWER than swap file!

(1) Another program may be editing the same file.  ...
(2) An edit session for this file crashed.
    ...

E5108: Error executing lua: Vim:E325: ATTENTION
stack traceback:
        [C]: in function 'bufload'
        .../harpoon/config.lua:110: in function 'select'
        .../harpoon/list.lua:182: in function 'select'
        ...
Press ENTER or type command to continue
```
This should be pretty simple to reproduce!?

After pressing enter, a blank buffer is shown. When trying `...:select(1)` once more, the file is loaded into the buffer and shown.
**Edit-1**: When trying to switch between `...:select(2)` and `...:select(1)`, the file is not shown anymore!?
**Edit-2**: The issue with Edit-1 only shows, when showing files with relative paths (added with `require("harpoon"):list():append()`).

This happens with neovim v0.9.4 and installed Lazy and harpoon2. I didn't try without Lazy.

Commenting out [line 110 in lua/harpoon/config.lua](https://github.com/ThePrimeagen/harpoon/blob/31701337377991c66eebb57ebd23ef01eb587352/lua/harpoon/config.lua#L110) seemed to work, at first. But it causes another issue, where the file/buffer is not loaded with `...:select(1)`, when preceded by `...:select(2)` (load a different file beforehand).
[...]

PR #498, opened 2024-01-20
Harpoon2 - fix for #444

kimabrandt-flx:
I've set `swapfile=false`, to let a test pass. It otherwise fails, when `swapfile` is enabled in the neovim-configuration.

I then added the fix for #444, which I've now testet and seems to be working!?

I didn't get around writing a test for it though! :\


kimabrandt-flx:
I've removed the `ok`-condition surrounding `vim.api.nvim_set_option_value("buflisted",...)`, to always apply it.

kimabrandt-flx:
Closing this PR, in favor of #504.

PR #500, opened 2024-01-23
Fix default list

kimabrandt-flx:
These changes try to address the issue with using `vim.fn.bufnr()` for loading buffers (see #503), and uses absolute paths instead of relative ones, for internal operations.


kimabrandt-flx:
Issue was addressed in #542 and fixed by [5b34471](https://github.com/ThePrimeagen/harpoon/commit/5b344710b7b3625435d639697375b2c3f715a94e).

PR #504, opened 2024-01-25
Fix for 444

kimabrandt-flx:
I modified my previous attempt of ignoring the swapfile-error from #444 in the tests.
Buffers are now being closed at the end of each test, so that they don't trigger the error in other tests.

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -108,6 +108,7 @@ function M.get_default_config()
                     bufnr = vim.fn.bufnr(list_item.value, true)
                 end
                 if not vim.api.nvim_buf_is_loaded(bufnr) then
+                    vim.api.nvim_set_current_buf(bufnr)
                     vim.fn.bufload(bufnr)
                     vim.api.nvim_set_option_value("buflisted", true, {
                         buf = bufnr,


lua/harpoon/config.lua
```diff
@@ -108,7 +108,7 @@ function M.get_default_config()
                     bufnr = vim.fn.bufnr(list_item.value, true)
                 end
                 if not vim.api.nvim_buf_is_loaded(bufnr) then
-                    vim.fn.bufload(bufnr)
+                    pcall(vim.fn.bufload, bufnr)
```
ThePrimeagen: we need to do something here if we cannot load the buffer

lua/harpoon/test/harpoon_spec.lua
```diff
@@ -41,14 +41,17 @@ describe("harpoon", function()
         }
 
         eq(expected, list.items)
+
+        vim.api.nvim_buf_delete(target_buf, { force = true })
```
ThePrimeagen: this should be cleaned up by the call to `be`.

utils should clear itself out and since these files were created via utils there should be bits keeping track there

ThePrimeagen:
remove the extra clean up code on the tests that _DO NOT_ use utils to create files (i think the config one is good because it doesn't use utils)

ThePrimeagen:
ok, i was thinking about this more.  if you want to move the `before_each` call to https://github.com/ThePrimeagen/harpoon/blob/harpoon2/lua/harpoon/test/utils.lua#L34 into a new function `after_each` i think that would be better.

that way we don't leave any hanging files ever (on last test).  its assumed that in the before_each we are in a clean env, and after each cleans up the env each time

kimabrandt-flx:
I've (again) changed my approach. I'm still learning!
The/my issue seems to be, that the tests are run in parallel!? At least my neovim seems to have `swapfile=true` set, when running the tests. So, I added a check, which should run the tests in sequence, when neovim has swapfiles enabled. I'm not sure that this is the right approach though!? But, it solves the issue with conflicting bufferes and prevents the swapfile-errors.
So, the file-cleanup doesn't seem to be necessary - my guess is - because the tests are run in their own neovim-instance - which is freeing resources when done - or something like that?! I'm not sure though!?
_Edit_: The tests also seem to succeed, without the call to `return_to_checkpoint()`, when either run sequentially or setting `vim.api.nvim_buf_set_option(bufnr, "swapfile", false)` inside the [`utils.create_file`-function](https://github.com/ThePrimeagen/harpoon/blob/a38be6e0dd4c6db66997deab71fc4453ace97f9c/lua/harpoon/test/utils.lua#L56)!?

lua/harpoon/config.lua
kimabrandt-flx: When searching for this error, I found [a similar issue in the vim-project](https://github.com/vim/vim/issues/8333). Bram was suggesting to ignore the error, which made me feel confident enough to report it here :)
I didn't see any issues with this change and it only gave an error (for me) when encountering a swapfile-message/-error!?
I'm sure there's a cleaner way of archiving the same!? It probably involves doing a check before calling `bufload`!?
I'll have another look at it. Also, I'll try to add a test for it.

kimabrandt-flx:
Here's a small fix, that seems to solve the swap-error!?

kimabrandt-flx:
[Here's a test](https://gist.github.com/kimabrandt-flx/8e7cb3d3f8cb620d6ffdc583e43f9ac9) that I've been working on. The issue with it is: that the _test-instance_ of neovim seems to behave different than _my instance_ of neovim, in respect to swapfiles!? The aforementioned "small fix" works with _my instance_, but not with the _test-instance_!?

kimabrandt-flx:
Issue was addressed in #430 and fixed by [747169f](https://github.com/ThePrimeagen/harpoon/commit/747169f07ae7b3b80ae0d655458265822788ac4f).

Issue #476, opened 2024-01-06
[Harpoon 2] Duplicate list items cause all following list items to be deleted


maxrzaw:
## Steps to reproduce
1. Mark a file
2. Go to a different file
3. Create a duplicate of the list item and then select it from the UI by hitting \<CR>
![image](https://github.com/ThePrimeagen/harpoon/assets/32819501/1e535e8d-de36-46fb-be58-24decbe4a36f)
4. Open UI again and select first item (now the file is opened)

I would expect this to take me to README.md on step 3, but it does not. I go to my back to the current file.

## Logs
config_default#create_list_item README.md
HarpoonList:append { index = -1, item = { context = { col = 6, row = 1 }, value = "README.md" } }
ui#toggle_quick_menu#opening __harpoon_files
select by keymap '<CR>' **<-- Does not work here**
ui#select_menu_item selecting item 2 from { "README.md", "README.md" } options
ui#close_menu name: __harpoon_files win and bufnr { bufnr = 17, win = 1010 }
toggle by BufLeave
ui#toggle_quick_menu#closing
ui#save { "README.md", "README.md" }
ui#toggle_quick_menu#opening __harpoon_files **<-- open up UI for second time**
select by keymap '<CR>' **<-- Works this time**
ui#select_menu_item selecting item 1 from { "README.md" } options
ui#close_menu name: __harpoon_files win and bufnr { bufnr = 11, win = 1005 }
toggle by BufLeave
ui#toggle_quick_menu#closing
ui#save { "README.md" }
config_default#select { context = { col = 31, row = 7 }, value = "README.md" } __harpoon_files

# Deletes items after the first duplicate

## Steps to Reproduce
Same as above, but there is another list item already. So when hitting \<CR>, the list looks like:
```
1 README.MD
2 README.MD
3 file1.txt
```
and my cursor is on the second README.md.
When I open the list back up, the list looks like:
```
1 README.MD
```
## Logs 2
ui#toggle_quick_menu#opening __harpoon_files
select by keymap '<CR>'
ui#select_menu_item selecting item 2 from { "README.md", "README.md", "file1.txt" } options
ui#close_menu name: __harpoon_files win and bufnr { bufnr = 4, win = 1002 }
toggle by BufLeave
ui#toggle_quick_menu#closing
ui#save { "README.md", "README.md", "file1.txt" }
config_default#create_list_item file1.txt
ui#toggle_quick_menu#opening __harpoon_files
select by keymap '<CR>'
[...]

PR #512, opened 2024-02-02
feat(picker): make the telescope harpoon-picker behave more like a builtin telescope file-picker

kimabrandt-flx:
This commit fixes the missing `entry.index` - in the marks-extension for telescope - which caused the index to always show `nil` - or rather `n…` - due to a fixed width of 2, in the dialogue.

This commit also added a dynamic width and left-padding for the index-column, so the index is being right-aligned.


kimabrandt-flx:
These changes add some enhancements for the telescope `Harpoon Marks`-picker.

rubiin (approved):

gvolpe:
This is great, thank you for putting up this work!

lua/telescope/_extensions/marks.lua
```diff
@@ -1,135 +1,284 @@
...
-    current_picker:refresh(generate_new_finder(), { reset_prompt = true })
+    current_picker:delete_selection(function() end)
+
+    return count
 end
 
-local move_mark_up = function(prompt_bufnr)
-    local selection = action_state.get_selected_entry()
-    local length = harpoon:list():length()
+local delete_mark_selections_prompt = function(prompt_bufnr)
```
gvolpe: nitpick: it would be great if this could be configurable (i.e. being able to disable this prompt and delete the mark straight away)

kimabrandt-flx: I could also imagine to override the mappings!? Something like this:

```lua
require("telescope").extensions.harpoon.marks(themes.get_dropdown({
    previewer = false,
    layout_config = { width = 0.6 },
    path_display = { truncate = 10 },
    attach_mappings = function(_, map)
        map("i", "<c-d>", hm_actions.delete_mark_selections)
        map("n", "<c-d>", hm_actions.delete_mark_selections)
        return true
    end,
}))
```

The different actions (e.g. `delete_mark_selections`, `delete_mark_selections_prompt`) would've to be exported first!?

For now, you could just override the mappings in the code and use `delete_mark_selections` instead of `delete_mark_selections_prompt`!

gvolpe: That would be even better! :)

Indeed, the actions need to be exported first. I'm using your fork now, thanks for doing this again 🙏🏽

ThePrimeagen:
If I can get a couple of approvals, I will merge this.

Not know the telescope stuff, so I need someone who can take the lead on reviewing these changes

lua/telescope/_extensions/marks.lua
kimabrandt-flx: You can now do:

```lua
local themes = require("telescope.themes")
local hm_actions = require("telescope._extensions.harpoon_marks.actions")
require("telescope").extensions.harpoon.marks(themes.get_dropdown({
    previewer = false,
    layout_config = { width = 0.6 },
    path_display = { truncate = 10 },
    attach_mappings = function(_, map)
        map("i", "<c-d>", hm_actions.delete_mark_selections)
        map("n", "<c-d>", hm_actions.delete_mark_selections)
        return true
    end,
}))
```

gvolpe (approved):
@ThePrimeagen I'm not a Telescope expert, but all I can say is this works perfectly! Thank you again @kimabrandt-flx 🙏🏽

ferjgm98 (approved):

11Firefox11:
Tried it out with [this](https://github.com/ThePrimeagen/harpoon/pull/512#discussion_r1552372809) code. I don't know telescope, I just tried it out and <a href="https://github.com/ThePrimeagen/harpoon/pull/512#issuecomment-2206108308"><s>it works well</s></a>. 

One thing I noticed though is that if you delete an item then the Harpoon indexes will change too, but in telescope the original numbers persist. For example in Telescope: 1, 2, 3, 4, 5 --delete 2nd--> 1, 3, 4, 5 (should be 1, 2, 3, 4 - which it will be if I close and reopen the whole thing).
Sadly I don't know how to update the currently displayed items in telescope so I can't implement this.

11Firefox11:
I used it for few days and it doesn't work well. The ordering really breaks when deleting items and then adding new ones. There will be empty rows in the original Harpoon window.

kimabrandt-flx:
@11Firefox11
The current Telescope marks-extension, has the same issue: https://github.com/ThePrimeagen/harpoon/issues/588
Maybe this could be fixed inside the [remove_at](https://github.com/ThePrimeagen/harpoon/blob/0378a6c428a0bed6a2781d459d7943843f374bce/lua/harpoon/list.lua#L205)-function!? Calling it, seems to leave a blank value (`nil`) inside the harpoon-list!? I'm not sure if the Telescope-picker should show the blank value, or if the blank value should be removed, inside the harpoon-list?!

I think - when restarting - you could be seeing something related to these issues!? https://github.com/ThePrimeagen/harpoon/issues/476, https://github.com/ThePrimeagen/harpoon/pull/573

PR #514, opened 2024-02-11
Get the current harpoon mark

skykosiner:
In harpoon one there was a neat little function you could call to get the current status.

So I added this feature to harpoon two. If you're currently in a buffer that is a mark when you call this function it will return the idx of the mark, or it will return nil.

The function can be called by doing
```lua
require("harpoon"):list():get_current_status()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,11 @@ harpoon:setup({
 
 ### Getting the current mark
 If you want to be able to get the current mark for something like your
-status line you just need to call `:lua require("harpoon"):list():get_current_status()"`.
+status line you just need to call:
+
+```lua
+require("harpoon"):list():get_current_status()
+```
 
 The function will return the idx of the current mark if your current buffer is
 a mark. Otherwise, it will return nil.
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ vim.keymap.set("n", "<C-S-N>", function() harpoon:list():next() end)
 
 ### Telescope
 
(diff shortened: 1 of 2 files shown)


jjcramer:
I'm not quite sure why it's failing the style test, but definitely looking forward to this being approved.

fsorodrigues:
Would love to see this bad boy implemented. The two cents i'd add is that:
```
vim.api.nvim_buf_get_name(0):gsub(vim.fn.getcwd() .. "/", "")
```
will fail miserably if you're anything like me and use `-` (the hyphen char) in your paths. Has to do with `-` being a "magical" character in Lua's pattern matching. I learned it the hard way.

So, you probably wanna escape that first (and other chars that need escaping). Could probably just use this implementation from Lua Núcleo: https://github.com/lua-nucleo/lua-nucleo/blob/v0.1.0/lua-nucleo/string.lua#L245-L267

skykosiner:
As it seems this isn't going to get merged. If you're interested in your lua configs or plugins or wherever you should be able to use this lua function to get the current mark.

```lua
function current_mark()
    local hasHarpoon = pcall(require, "harpoon")

    if not hasHarpoon then
        return ""
    end

    local current_file = vim.api.nvim_buf_get_name(0):gsub(vim.fn.getcwd() .. "/", "")
    local marks = require("harpoon"):list().items

    for idx, item in ipairs(marks) do
        if item.value == current_file then
            return tostring(idx)
        end
    end

    return ""
end
```

PR #518, opened 2024-02-19
ci: generate vimdocs on push

davvid:
Use [panvimdoc](https://github.com/kdheepak/panvimdoc) to generate vim docs on push.

Add a `make doc` target for local debugging.

Closes: #401


davvid:
shoot I didn't see #432, closing

PR #525, opened 2024-02-29
config: use mkview and loadview to restore marks and cursors

davvid:
Vim has a builtin mechanism for saving and restoring cursor positions and marks across restarts. Leverage mkview and loadview to make harpoon restore marks and cursor positions after quitting.

https://www.reddit.com/r/vim/comments/ex1vlv/comment/fg63erw

Closes: #441

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -100,7 +100,7 @@ function M.get_default_config()
                 if list_item == nil then
                     return
                 end
-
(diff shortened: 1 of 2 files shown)


davvid:
Closing per the discussion in #441 as this was primarily a proof of concept demo. We should make it so that harpoon restores cursor positions itself without all of the mkview / loadview overhead.

PR #533, opened 2024-03-11
list: save and restore cursor positions

### LINKED ISSUE #441: Harpoon doesn't save cursor position after quit

davvid:
Harpoon was not remembering the cursor position after quitting nvim.
Ensure that the cursor information is always saved and applied so that
we get back to exactly where we last were in the harpooned file.

Closes: #441

### DIFF
--- a/lua/harpoon/list.lua
+++ b/lua/harpoon/list.lua
@@ -1,5 +1,4 @@
 local Logger = require("harpoon.logger")
-local utils = require("harpoon.utils")
 local Extensions = require("harpoon.extensions")
 
 local function guess_length(arr)
@@ -259,7 +258,7 @@ function HarpoonList:resolve_displayed(displayed, length)
     for i = 1, length do
         local v = displayed[i]
         local index = index_of(list_displayed, self._length, v)
-        if utils.is_white_space(v) then
+        if Utils.is_white_space(v) then
             new_list[i] = nil
         elseif index == -1 then
             new_list[i] = self.config.create_list_item(self.config, v)
--- a/lua/harpoon/list.lua
+++ b/lua/harpoon/list.lua
@@ -1,5 +1,6 @@
 local Logger = require("harpoon.logger")
 local Extensions = require("harpoon.extensions")
+local Utils = require("harpoon.utils")
 
 local function guess_length(arr)
     local last_known = #arr
@@ -284,6 +285,17 @@ function HarpoonList:resolve_displayed(displayed, length)
     end
 end
 
+function HarpoonList:sync_cursor()
+    local current_path = Utils.normalize_path(
+        vim.api.nvim_buf_get_name(vim.api.nvim_get_current_buf()),
+        self.config.get_root_dir()
+    )
+    local item = self:get_by_value(current_path)
+    if item then
+        HarpoonList._sync_cursor(item)
+    end
+end
+
 function HarpoonList:select(index, options)
     local item = self.items[index]
     if item or self.config.select_with_nil then
@@ -361,6 +373,15 @@ function HarpoonList.decode(list_config, name, items)
     end
 
     return HarpoonList:new(list_config, name, list_items)
+end
+
+--- @return integer[]
+--- @param item HarpoonItem
+function HarpoonList._sync_cursor(item)
+    local pos = vim.api.nvim_win_get_cursor(0)
+    item.context.row = pos[1]
+    item.context.col = pos[2]
+    return pos
 end
 
 return HarpoonList
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -1,5 +1,6 @@
 local Extensions = require("harpoon.extensions")
 local Logger = require("harpoon.logger")
+local List = require("harpoon.list")
 local Utils = require("harpoon.utils")
 
 local function to_exact_name(value)
@@ -102,13 +103,11 @@ function M.get_default_config()
                     return
                 end
 
+                list:sync_cursor()
                 options = options or {}
 
                 local bufnr = vim.fn.bufnr(to_exact_name(list_item.value))
-                local set_position = false
                 if bufnr == -1 then -- must create a buffer!
-                    set_position = true
-                    -- bufnr = vim.fn.bufnr(list_item.value, true)
                     bufnr = vim.fn.bufadd(list_item.value)
                 end
                 if not vim.api.nvim_buf_is_loaded(bufnr) then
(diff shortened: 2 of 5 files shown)


romado77:
@ThePrimeagen could you please review this PR? It addresses a long-awaited fix.

maxrzaw:
This seems like an improvement, but doesn't appear to work unless you first navigate to the file using harpoon. I have tested this with sync_on_ui_close both true and false.

For example:
1. vim A.txt
2. Add mark on line 5
3. Quit
4. vim A.txt
5. Move to line 10
6. Quit
7. vim B.txt
8. navigate to A.txt in harpoon
9. Should be on line 10 and not line 5


Adding this autocmd appears to fixe this issue with the default config.
```lua
        vim.api.nvim_create_autocmd({ "QuitPre" }, {
            pattern = "*",
            callback = function()
                local bufnr = vim.api.nvim_get_current_buf()
                local path = vim.api.nvim_buf_get_name(bufnr)
                for _, it in ipairs(Harpoon:list().items) do
                    local value = it.value
                    if value == path then
                        Harpoon:list():append() -- now Harpoon:list():add()
                        break
                    end
                end
            end,
        })
```

However, I still have an issue with the position not updating when switching between files (not using harpoon) and then navigating to the file using harpoon. Easy example of this is make a mark, move down, go to alternate file, go back using harpoon and you will be where the mark is and not where you last left off. I guess this is a different issue since it isn't while quitting, but it seems similar.

ThePrimeagen:
I have done a lot of work on getting cursors to save constantly. The latest update does a distributed file style based on settings key, this allows saving upon every change 

I would first make sure the latest harpoon2 still has the issue you're looking for

maxrzaw:
@ThePrimeagen I just checked with latest, harpoon2 does still have the issue.

I was able to update the changes @davvid made to work with the latest harpoon2. Those changes along with the now much simpler autocmd seem to be working.
```lua
        vim.api.nvim_create_autocmd({ "QuitPre" }, {
            pattern = "*",
            callback = function()
                -- Do this for all your lists
                Harpoon:list():sync_cursor()
            end,
        })
```
I tried putting it in harpoon/init.lua and using _for_each_list, but it did not work.
The problem is that _for_each_list does not run callbacks for any lists when I open up a file, move around, but never interact with harpoon at all.

my updated branch https://github.com/maxrzaw/harpoon/tree/sync-cursor seems to be working. There might be some redundancies with the combination of the new changes to harpoon2 and the changes from this PR.

davvid:
Thanks for the heads-up @maxrzaw -- I've rebased this branch and updated to the latest harpoon2.

The out-of-bounds cursor tests had to be updated with these changes, which I suspect is okay since the results are getting updated to match the file bounds. That's probably worth another look.

@maxrzaw is the expectation that users will need to install the `QuitPre` autocmd themselves if they want a top-to-bottom fix?

The behavior is better for more scenarios with these changes at least. w/out these changes the cursor position is lost even when switching between two harpooned files. For example, move around in one file, harpoon to a 2nd file, and then harpoon back and the cursor will be in a different position from whence we came. It's in the correct position when using this branch.

Things `:bNext` and `:bPrev` do seem to reset the cursor, though, but at least this is better than it was before.

Curious, which scenario does the autocmd handle? I don't see any change in behavior when using `save-cursor-position` w/ and w/out the autocmd, though maybe that's because I have these options set?

```lua
        opts = {
            settings = {
                save_on_toggle = true,
                sync_on_ui_close = true,
            },
        },
```

romado77:
> I have done a lot of work on getting cursors to save constantly. The latest update does a distributed file style based on settings key, this allows saving upon every change
> 
> I would first make sure the latest harpoon2 still has the issue you're looking for

For me the position isn't being saved when I quit, regardless of whether I save the buffer or not. The only way I've found to make it work is by toggling the Harpoon list again (leader + h), selecting the same file from the list that's already open, and then quitting. Doing this saves the current position, allowing me to jump back to it when quitting Nvim and reopening the file via Harpoon. All other solutions don't work including AutoCmd suggested by @maxrzaw.
BTW, I'm on the latest commit.

maxrzaw:
> > I have done a lot of work on getting cursors to save constantly. The latest update does a distributed file style based on settings key, this allows saving upon every change
> > I would first make sure the latest harpoon2 still has the issue you're looking for
> 
> For me the position isn't being saved when I quit, regardless of whether I save the buffer or not. The only way I've found to make it work is by toggling the Harpoon list again (leader + h), selecting the same file from the list that's already open, and then quitting. Doing this saves the current position, allowing me to jump back to it when quitting Nvim and reopening the file via Harpoon. All other solutions don't work including AutoCmd suggested by @maxrzaw. BTW, I'm on the latest commit.

Latest commit of which branch? I was using latest on harpoon2 with additional changes from this branch. https://github.com/maxrzaw/harpoon/tree/sync-cursor to be exact.

maxrzaw:
> Thanks for the heads-up @maxrzaw -- I've rebased this branch and updated to the latest harpoon2.
> 
> 
> 
> The out-of-bounds cursor tests had to be updated with these changes, which I suspect is okay since the results are getting updated to match the file bounds. That's probably worth another look.
> 
> 
> 
> @maxrzaw is the expectation that users will need to install the `QuitPre` autocmd themselves if they want a top-to-bottom fix?
> 
> 
> 
> The behavior is better for more scenarios with these changes at least. w/out these changes the cursor position is lost even when switching between two harpooned files. For example, move around in one file, harpoon to a 2nd file, and then harpoon back and the cursor will be in a different position from whence we came. It's in the correct position when using this branch.
> 
> 
> 
> Things `:bNext` and `:bPrev` do seem to reset the cursor, though, but at least this is better than it was before.
> 
> 
> 
> Curious, which scenario does the autocmd handle? I don't see any change in behavior when using `save-cursor-position` w/ and w/out the autocmd, though maybe that's because I have these options set?
> 
> 
> 
> ```lua
> 
>         opts = {
> 
>             settings = {
> 
>                 save_on_toggle = true,
> 
>                 sync_on_ui_close = true,
> 
>             },
> 
>         },
> 
> ```


These steps:
1. vim A.txt
2. Add mark on line 5
3. Quit
4. vim A.txt
5. Move to line 10
6. Quit
7. vim B.txt
8. navigate to A.txt in harpoon
9. Should be on line 10 and not line 5

The key is that I open a harpooned file and move the cursor, but do nothing at all to interact with harpoon, then quit, this results in _for_each_list not working as expected.

ThePrimeagen:
Let me double check everything, I think I know what happened. 

I should get some time this week to go over everything

romado77:
> > > I have done a lot of work on getting cursors to save constantly. The latest update does a distributed file style based on settings key, this allows saving upon every change
> > > I would first make sure the latest harpoon2 still has the issue you're looking for
> > 
> > 
> > For me the position isn't being saved when I quit, regardless of whether I save the buffer or not. The only way I've found to make it work is by toggling the Harpoon list again (leader + h), selecting the same file from the list that's already open, and then quitting. Doing this saves the current position, allowing me to jump back to it when quitting Nvim and reopening the file via Harpoon. All other solutions don't work including AutoCmd suggested by @maxrzaw. BTW, I'm on the latest commit.
> 
> Latest commit of which branch? I was using latest on harpoon2 with additional changes from this branch. https://github.com/maxrzaw/harpoon/tree/sync-cursor to be exact.

harpoon2

romado77:
> > > I have done a lot of work on getting cursors to save constantly. The latest update does a distributed file style based on settings key, this allows saving upon every change
> > > I would first make sure the latest harpoon2 still has the issue you're looking for
> > 
> > 
> > For me the position isn't being saved when I quit, regardless of whether I save the buffer or not. The only way I've found to make it work is by toggling the Harpoon list again (leader + h), selecting the same file from the list that's already open, and then quitting. Doing this saves the current position, allowing me to jump back to it when quitting Nvim and reopening the file via Harpoon. All other solutions don't work including AutoCmd suggested by @maxrzaw. BTW, I'm on the latest commit.
> 
> Latest commit of which branch? I was using latest on harpoon2 with additional changes from this branch. https://github.com/maxrzaw/harpoon/tree/sync-cursor to be exact.

harpoon2

> On branch harpoon2
> Your branch is up to date with 'origin/harpoon2'.
>
> nothing to commit, working tree clean

lua/harpoon/list.lua
```diff
@@ -285,6 +285,17 @@ function HarpoonList:resolve_displayed(displayed, length)
     end
 end
 
+function HarpoonList:sync_cursor()
+    local current_display = Utils.normalize_path(
```
maxrzaw: Calling this `current_display` is a bit misleading as it is used in `self:get_by_value(current_display)`.

davvid: Any sugs for what we might want to name for this variable instead? What do you think about  `displayed_path`?

maxrzaw: I would go with `current_path`.

Issue #540, opened 2024-03-24
The name is ....... sometimes too broad for a buffer


abeldekat:
https://github.com/ThePrimeagen/harpoon/assets/58370433/385fddb4-41ab-4491-840c-5cd0e22a2b77

Two files are harpooned:

- `namespace/init.lua`
- `init.lua`

Steps: 
1. Open neovim and `harpoon select` to item one.
2. Now, `harpoon select` to item two. Observe that item two is **not** opened.

Explanation:

The default `select` function in `harpoon/config`, line 104

```lua
local bufnr = vim.fn.bufnr(list_item.value)
local set_position = false
if bufnr == -1 then
    set_position = true
    bufnr = vim.fn.bufnr(list_item.value, true)
end
```

Function `vim.fn.bufnr`, given the value `init.lua` in step 2, will return a valid bufnr: The buffer belonging to `namespace/init.lua`!. See `:h bufname` and `:h bufnr`.

<details>
<summary>Repro setup</summary>

```txt
Use:
  mkdir ~/.config/repro
  cd ~/.config/repro
  NVIM_APPNAME=repro nvim init.lua
Remove:
  rm -rf ~/.local/share/repro ~/.local/state/repro ~/.local/cache/repro
  rm -rf ~/.config/repro
```

</details>

<details>
<summary>Repro init.lua</summary>

```lua

local function clone(path_to_site)
  local mini_path = path_to_site .. "pack/deps/start/mini.deps"
  if not vim.loop.fs_stat(mini_path) then
    vim.cmd('echo "Installing `mini.deps`" | redraw')
    local clone_cmd =
      { "git", "clone", "--filter=blob:none", "https://github.com/echasnovski/mini.deps", mini_path }
    vim.fn.system(clone_cmd)
    vim.cmd("packadd mini.deps | helptags ALL")
    vim.cmd('echo "Installed `mini.deps`" | redraw')
  end
end

local path_to_site = vim.fn.stdpath("data") .. "/site/"
clone(path_to_site)
local MiniDeps = require("mini.deps")
MiniDeps.setup({ path = { package = path_to_site } })
local add, now = MiniDeps.add, MiniDeps.now

vim.g.mapleader = " "
now(function()
  add({ source = "ThePrimeagen/harpoon", checkout = "harpoon2", depends = { "nvim-lua/plenary.nvim" } })
  local Harpoon = require("harpoon")
  Harpoon:setup({ settings = { save_on_toggle = true } })

[...]

PR #541, opened 2024-03-24
fix: The name is ....... sometimes too broad for a buffer

abeldekat:
Fixes issue: #540 

Related: This [PR](https://github.com/ThePrimeagen/harpoon/pull/503), regarding tests.

To consider:

1. Maybe there are more places where something similar happens
2. Maybe the question "is there a buffer with this exact name" should be asked to `harpoon.buffer.lua`
3. I could not get a test going. Perhaps, this is more an e2e test, needing a `testdata` folder, with 2 files and a harpoon.json

Non-failing test:

```lua
it("can handle the same filename in different directories", function()
    local file_one = "/tmp/namespace/init.lua"
    local buf_one = utils.create_file(file_one, { "foo", "bar" }, 1, 0)
    harpoon:list():append()

    local file_two = "/tmp/init.lua"
    local buf_two = utils.create_file(file_two, {
        "baz",
        "qux",
    }, 1, 0)
    harpoon:list():append()

    harpoon:sync()
    local list = harpoon:list()

    eq(list.items, {
        { value = file_one, context = { row = 1, col = 0 } },
        { value = file_two, context = { row = 1, col = 0 } },
    })

    vim.cmd("bwipeout! " .. buf_one)
    vim.cmd("bwipeout! " .. buf_two)

    harpoon:list():select(1)
    eq("/tmp/namespace/init.lua", vim.api.nvim_buf_get_name(0))
    harpoon:list():select(2)
    eq("/tmp/init.lua", vim.api.nvim_buf_get_name(0))
end)
```

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -4,6 +4,9 @@ local Path = require("plenary.path")
 local function normalize_path(buf_name, root)
     return Path:new(buf_name):make_relative(root)
 end
+local function to_exact_name(value)
+    return "^" .. value .. "$"
+end
 
 local M = {}
 local DEFAULT_LIST = "__harpoon_files"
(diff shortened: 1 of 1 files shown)


rubiin (approved):

ThePrimeagen:
i can merge this, hodl

i hope this cures these weird file issues*

PR #543, opened 2024-04-01
Fix indexing and use display configuration for telescope picker

rrothenberger:
Currently, index in telescope is not displayed properly:
image

As well as configuration for `display` doesn't change how entries are displayed, this fixes both issues

### DIFF
--- a/lua/telescope/_extensions/marks.lua
+++ b/lua/telescope/_extensions/marks.lua
@@ -6,47 +6,36 @@ local pickers = require("telescope.pickers")
 local conf = require("telescope.config").values
 local harpoon = require("harpoon")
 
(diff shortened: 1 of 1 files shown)


lua/telescope/_extensions/marks.lua
```diff
@@ -6,47 +6,45 @@ local pickers = require("telescope.pickers")
 local conf = require("telescope.config").values
 local harpoon = require("harpoon")
 
-local function filter_empty_string(list)
```
rrothenberger, line 9: I removed that, because if you add invalid entry it would be filtered out from telescope view and you won't be able to deleted it from harpoon, even if it is there and is taking up index

PR #544, opened 2024-04-01
feat(list-api): expose index for later usage

radvil:
- I have a custom append/prepend list where I need to notify the user if current file is already in the list, if it is, I want to hint the use for the exact index, so we can go faster

### DIFF
--- a/lua/harpoon/list.lua
+++ b/lua/harpoon/list.lua
@@ -126,7 +126,7 @@ function HarpoonList:get_by_display(name)
     if index == -1 then
         return nil
     end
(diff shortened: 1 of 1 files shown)


radvil:
Here is my current preview to hint my brain
[Screencast_20240402_004445.webm](https://github.com/ThePrimeagen/harpoon/assets/36059968/03052f73-1997-4994-bab9-3b36b4ba5e59)

ThePrimeagen:
very cool plugin and yes, i can support this behavior

PR #554, opened 2024-04-04
Add DirChanged autocmd

### LINKED ISSUE #553: (bug) Opening the quickmenu after starting vim from '~/' then `:cd` to a dir deletes the list at that path
**WARNING**

---------------
## Steps:
1. Open nvim from ~/some/path/
2. Call `harpoon:list():add()` on a file
3. Open the harpoon menu, see addition
3. `:cd` (to home dir, or anything)
4. Open harpoon menu, see nothing
5. Close menu.
6. `:q!`
7. Reopen vim in same path
8. Open harpoon menu, see nothing

Makaze:
Resolves #553

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -169,6 +169,14 @@ function Harpoon.setup(self, partial_config)
             end,
         })
 
+        vim.api.nvim_create_autocmd({ "DirChanged" }, {
+            group = HarpoonGroup,
+            callback = function()
+                require("harpoon").logger:log("reset by DirChanged")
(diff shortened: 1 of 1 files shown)


Makaze:
This bug only occurs with ~. Prime closed the issue, but for posterity, the bug occurs when you open nvim from user home, then `:cd` to desired dir from within neovim. Harpoon does not behave as expected in that specific case: it deletes the list at the path you `:cd` to.

[asciicast](https://asciinema.org/a/cacy2KkVxLl2SJlSCXRSfN5Ar)

PR #562, opened 2024-04-06
fix(ui_nav_wrap): fix cycling with `list:next()` and `list:prev()`

kimabrandt-flx:
Allow the UI to cycle through the marks, by default. To disable cycling, pass `ui_nav_wrap=false`:

```lua
require("harpoon"):list():prev({ ui_nav_wrap = false })
require("harpoon"):list():next({ ui_nav_wrap = false })
```

This was mentioned in #461 and #341.


ThePrimeagen:
This is opt out cycling

Will not merge as it breaks behavior

PR #569, opened 2024-04-09
Fix1

ThePrimeagen:
@sug44 here you go

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -4,9 +4,6 @@ local Path = require("plenary.path")
 local function normalize_path(buf_name, root)
     return Path:new(buf_name):make_relative(root)
 end
-local function to_exact_name(value)
-    return "^" .. value .. "$"
-end
 
 local M = {}
 local DEFAULT_LIST = "__harpoon_files"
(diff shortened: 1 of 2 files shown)


ThePrimeagen:
a couple of breaking tests... looking

ThePrimeagen:
@sug44 removing exact path has caused bugs

sug44:
the tests were breaking on the parent commit as well, so i didnt try to fix any of them

ThePrimeagen:
yeah, but the removal has caused somethings

there is also the usage of realpath that is causing breaking

ThePrimeagen:
also... i realized... i cannot have this change

i am going to have to think of something different.

this change breaks one of the core concepts of harpoon.  i can have 2 repos, both with the same `key()` in settings (think workspaces) and it will return the same results.  i _must_ have relative paths

sug44:
Yeah i see now that something changed, I'll investigate. About the https://github.com/ThePrimeagen/harpoon/pull/568#discussion_r1558095717 comment. Yeah it looks wrong because its not apperent what `a` and `b` is, but the purpose of this `index_of` is to find a item from the list that corresponds with the given path, so `a` is always absolute path and `b` is always `listItem.value`. I think I'll just add some comments so it doesnt look so scary

PR #571, opened 2024-04-10
fix: #570 native swap file error suppressed

### LINKED ISSUE #570: [harpoon2] Native swap file error prompt suppressed by harpoon
**Current Behavior**
When calling `harpoon:list():select(n)` a file fails to open if there is an existing swap file. The native confirmation prompt works well here, but it gets suppressed by harpoon and some of its contents are shown to the user without the ability to select any of the usual options ([O]pen Read-Only, [E]dit Anyway, [R]ecover, [Q]uit, [A]bort).

When the user presses Enter or any other key, the dialog disappears but the file does not open.
Running the initial command (`harpoon:list():select(n)`) a second time opens the file but does not inform the user of the existing swap file. The swap file is never modified or removed so the problem persists until the user either manually deletes the swap file, force saves the file (`:w!`) or opens the file using a native command (`:e`) and interacts with the native prompt to resolve the swap.
Screenshot from 2024-04-10 15-51-44

**Expected Behavior**
Essentially, it should feel the same as opening the file natively with `:e`
If there is an existing swap file the user should be presented with the native confirmation prompt (or equivalent) and have the ability to press `O`, `E`, `R`, `Q`, or `A` to perform one of the actions Neovim normally presents as a resolution.
Screenshot from 2024-04-10 15-51-29

**Setup**
I started with a basic kickstart.nvim and added harpoon as the only user plugin. No other changes were made prior to the test.
I then opened the same file in two different terminals and experienced the issue in the second terminal.

dabstractor:
Closes #570

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -114,7 +114,7 @@ function M.get_default_config()
                     bufnr = vim.fn.bufadd(list_item.value)
                 end
                 if not vim.api.nvim_buf_is_loaded(bufnr) then
-                    vim.fn.bufload(bufnr)
+                    vim.api.nvim_command("buffer " .. bufnr)
                     vim.api.nvim_set_option_value("buflisted", true, {
                         buf = bufnr,
                     })


kimabrandt-flx:
@dabstractor Did you try to get rid of the red-colored exception, after the dialog? I like this solution better than the one suggested in [comment](https://github.com/ThePrimeagen/harpoon/issues/444#issuecomment-1920000837), if the exception went away :)

dabstractor:
@kimabrandt-flx The 2nd screenshot I posted with the red error message was after opening the file with the default built-in file browser. This is the default Neovim experience. So even though it's obviously an eyesore I don't believe there's anything to "fix" per se. This is just how Neovim behaves.

I do believe this operation requires proper error-handling, though, in order to prevent opening the file with an existing swap if the user chooses to [Q]uit or [A]bort from this point, but my gut tells me that getting rid of an error produced by Neovim's default experience is probably not a job for Harpoon.

Please let me know if I misunderstood your question, it's certainly possible I did.

kimabrandt-flx:
@dabstractor

I think you're right, that there's not much to do, about the swap-error!

I think with never versions of Neovim (e.g. v0.10.0), this issue will go away!? See https://github.com/neovim/neovim/pull/25336!

Having said that; I was referring to, when I try this change with Neovim v0.9.5.
```
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1702233742
```
After choosing one of: `[O]pen`, `(E)dit`, `(R)ecover`, `(Q)uit`, `(A)bort`, I get the `E5108: Error executing lua`, together with the `E325: ATTENTION`.

Also, choosing `(Q)uit` or `(A)bort` works as expected here: The file is not loaded.

I was just wondering, if there was a way to suppress the error in Lua!? But, never mind! I'm already using v0.10.0 for daily activities ;)

Edit:
I guess, wrapping `vim.api.nvim_command()` with a `pcall` works!? But, will take some more changes!?

dabstractor:
I like your PR a lot better, thank you.

PR #572, opened 2024-04-11
fix #565: setup data after config loaded

### LINKED ISSUE #565: [harpoon2] when custom key is specified, persistent not working correctly
**WARNING**

---------------

My harpoon setup:
```
harpoon:setup({
    settings = {
        save_on_toggle = true,
        sync_on_ui_close = true,
        key = function()
            local key = vim.fn.getcwd()
            local current_branch = utils.get_current_branch()
            if current_branch ~= nil then
                key = string.format("%s#%s", key, current_branch)
            end
            return key
        end
    }
})
```

I add and remove file to harpoon, but when i close and open nvim again, my harpoon list is empty.

perseus-me:


akyrey:
@rusnasonov  thank you for the PR.

What about leaving the data initialization and `sync_on_change` function also inside `Harpoon::new()` and edit `Harpoon::setup()` like you did but with `sync_on_change(self)` moved right before the final return?

perseus-me:
> What about leaving the data initialization and `sync_on_change` function also inside `Harpoon::new()`

It's creates additional empty data file with default key, which will never be used. If it's ok - why not.

> but with sync_on_change(self) moved right before the final return?

Ok

maxrzaw:
This also fixes #577

lua/harpoon/init.lua
```diff
@@ -146,6 +146,7 @@ function Harpoon.setup(self, partial_config)
 
     ---@diagnostic disable-next-line: param-type-mismatch
     self.config = Config.merge_config(partial_config, self.config)
+    self.data = Data.Data:new(self.config)
```
iserpent, line 149: Found this MR just before creating my pull request :) Looks like it takes time to get it merged.
One suggestion from me would be to check at least for the `partial_config` to be non-empty. Otherwise we don't need to reload data and  calling `sync_on_change` (e.g. if it was just harpoon.setup({})).
```suggestion
    if partial_config then
        self.data = Data.Data:new(self.config)
    end
```
(and the same `if` before calling `sync_on_change`).

PR #573, opened 2024-04-11
fix(encode): use a `for`-loop, instead of `next()` or `ipairs()`, to prevent loss of data

kimabrandt-flx:
This fixes the issue, when there's an empty line inside the quick-menu, which causes everything after the blank line to be forgotten (after Neovim is restarted).

This also restores the blank lines, which I'm not certain of is expected or not?!

### DIFF
--- a/lua/harpoon/list.lua
+++ b/lua/harpoon/list.lua
@@ -344,8 +344,11 @@ end
 --- @return string[]
 function HarpoonList:encode()
     local out = {}
-    for _, v in ipairs(self.items) do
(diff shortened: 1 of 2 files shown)


kimabrandt-flx:
There still seems to be an issue, when there are many blank items!? Some of them are just removed!? But, non-nil items seem to get restored alright!?

PedroBinotto:
I think I might have found/fixed the same issue here #627 #628

PedroBinotto:
Pretty sure this was fixed in #628. Should we close this issue?

PR #574, opened 2024-04-11
feat(index-tracking): keep track of the `list._index` while navigating

kimabrandt-flx:
This change tries to keep track of the `list._index`, by using an `autocmd` for `BufEnter` - which updates the index when entering a harpooned file - and setting the index for different functions inside the list.

There's room for improvements, but maybe it can help!?

This also removes the `autocmd Filetype harpoon` - which didn't seem to highlight the row?! - and replaces it with passing the `index` for setting the cursor to a row.

Maybe there should be an user-option, to enable this?!

This also merges #573, since I felt it's somehow related.

### DIFF
--- a/lua/harpoon/buffer.lua
+++ b/lua/harpoon/buffer.lua
@@ -29,18 +29,6 @@ end
 
 ---@param bufnr number
 function M.setup_autocmds_and_keymaps(bufnr)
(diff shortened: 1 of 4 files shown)


LehMichael:
Works well for me. Would be nice, if it could get merged.

PR #575, opened 2024-04-13
fix(swapfile): ignore the swapfile-message

kimabrandt-flx:
This change supresses the `E325: ATTENTION`-message - when a swapfile is encountered - by catching the error and stop execution when the user chooses one of `(R)ecover`, `(Q)uit` or `(A)bort`.

This addresses the issues https://github.com/ThePrimeagen/harpoon/issues/444 (?) and https://github.com/ThePrimeagen/harpoon/issues/570.

This duplicates PR #571.

### DIFF
--- a/lua/harpoon/config.lua
+++ b/lua/harpoon/config.lua
@@ -114,7 +114,17 @@ function M.get_default_config()
                     bufnr = vim.fn.bufadd(list_item.value)
                 end
                 if not vim.api.nvim_buf_is_loaded(bufnr) then
-                    vim.fn.bufload(bufnr)
+                    local abort = false
+                    xpcall(vim.api.nvim_command, function(e)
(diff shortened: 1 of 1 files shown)


kimabrandt-flx:
Only problem I see is, that the rest of the function - set position and emit event - is not getting executed, when doing a `(R)ecover` :\

midgramr:
Hey, why'd you close this? Would be nice to not get these warnings

kimabrandt-flx:
> Hey, why'd you close this? Would be nice to not get these warnings

[Newer Neovim versions](https://neovim.io/doc/user/news-0.10/) handle the swap-warning much more graceful.
Also, I realized it's not polite to open PRs left and right, before talking about the issue beforehand.
Also, I simply use marks in Neovim now :)

midgramr:
Hmm, yeah I guess global marks are basically harpoon buffers but more granular

PR #576, opened 2024-04-15
feat: added viewports

lars-vc:
Solving #417
This allows for viewports to be saved when switching files with harpoon.
Important note, I didn't understand how to access the settings table within the select function, so this code isn't triggered until that line is changed.
Could you tell me how to properly access the user's settings?


maxrzaw:
This looks pretty cool. I think the `restore_viewport` setting would make more sense if it was a setting for a specific list rather than for harpoon. Similar to `select_with_nil`. This would then allow you to access this setting during `select`.

PR #598, opened 2024-05-19
fix: initialize data during Harpoon.setup

### LINKED ISSUE #577: Harpoon doesn't remember my marks after close
## I get the issue only on commits 3e32576076a7897eb806f1437cac928ac60d47c3 and later.

Harpoon works just fine while nvim is open, but when I close nvim, it forgets the marks. I'm guessing this commit's hashing thing doesn't work for me.

## Temporary solution for users: 
Pin your version to e76cb03 in your plugin manager of choice. For lazy, just add ```commit = 'e76cb03'```, then run ```:Lazy sync```. If it doesn't work right away, you can restart nvim and it should work now.

maxrzaw:
fixes #577 

Since `Data:new()` was being called before `Harpoon.setup(config)`, it was loading the file for the hash of `vim.loop.cwd()` rather than key. This causes the data file for key to be overwritten when sync is called.

I am having trouble running the tests on a Mac so I have no idea if they are passing.

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -147,6 +147,7 @@ function Harpoon.setup(self, partial_config)
     ---@diagnostic disable-next-line: param-type-mismatch
     self.config = Config.merge_config(partial_config, self.config)
     self.ui:configure(self.config.settings)
+    self.data = Data.Data:new(self.config)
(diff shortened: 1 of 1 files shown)


maxrzaw:
Oops, this is a duplicate of #572

PR #601, opened 2024-05-28
Added instruction on how to remove an entry from the quick menu

ar7casper:


MattTaylorQuodd:
I think you should note that you also must write the buffer `:w` and not just close the menu.

ar7casper:
@MattTaylorQuodd Yep, you're right, will add, thanks.

PR #614, opened 2024-07-02
fix: properly handle directory changes

### LINKED ISSUE #612: Harpoon2: Wrong state is loaded when working with multiple tabs - having different working dirs
**WARNING**

---------------

When opening a tab, and setting a different working folder in the new tab, the harpoon list is not loaded next time I start neovim.

I believe that the issue occurs because the data file uses the hash of the current working directory for to generate the file name.

```
local function fullpath(config)
    local h = hash(filename(config))
    return string.format("%s/%s.json", data_path, h)
end
```

When neovim loads, it uses the hash of the initial working folder, but when I open a new tab with a different working folder, it appears that the hash of the new working folder is used when saving the lists. So now lists are saved in a different file than the one loaded at startup.

Is there a reason for the hashing rather than one file? 

The data file already contains data for multiple working directories already.

cosmicboots:
This change clears out the in-memory harpoon data and re-read it from the disk whenever a `DirChanged` event is triggered. This makes sure the in memory data always reflects the harpoon lists for the current working directory.

Fixes #612

### DIFF
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -100,6 +100,7 @@ end
 ---Resets harpoon data by reading from disk
 function Harpoon:reset()
     self.data = Data.Data:new(self.config)
+    self.lists = {}
 end
 
 function Harpoon:sync()
--- a/lua/harpoon/init.lua
+++ b/lua/harpoon/init.lua
@@ -157,25 +157,28 @@ function Harpoon.setup(self, partial_config)
     ---TODO: should we go through every seen list and update its config?
 
     if self.hooks_setup == false then
-        vim.api.nvim_create_autocmd({ "BufLeave", "VimLeavePre", "DirChanged" }, {
-            group = HarpoonGroup,
-            pattern = "*",
-            callback = function(ev)
-                if ev.event == "DirChanged" then
-                    self:reset()
-                end
-                self:_for_each_list(function(list, config)
-                    local fn = config[ev.event]
-                    if fn ~= nil then
-                        fn(ev, list)
+        vim.api.nvim_create_autocmd(
+            { "BufLeave", "VimLeavePre", "DirChanged" },
+            {
+                group = HarpoonGroup,
+                pattern = "*",
+                callback = function(ev)
+                    if ev.event == "DirChanged" then
+                        self:reset()
                     end
+                    self:_for_each_list(function(list, config)
+                        local fn = config[ev.event]
+                        if fn ~= nil then
+                            fn(ev, list)
+                        end
 
-                    if ev.event == "VimLeavePre" then
-                        self:sync()
-                    end
-                end)
-            end,
-        })
+                        if ev.event == "VimLeavePre" then
+                            self:sync()
+                        end
+                    end)
+                end,
+            }
+        )
 
         self.hooks_setup = true
     end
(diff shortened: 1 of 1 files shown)


pcyman:
Tested this and it fixes #612 for me.
Hope it gets merged!

stroiman:
I finally got around to test this, and based on the initial testing, it works as expected. (I submitted #612)

E.g. I have keyboard shortcuts to open my init.lua in either a new tab, or new window, and set the local dir for better file navigation - e.g. telescope find files show either project-related files or files in my modularised configuration (`:tabnew +tcd\ %:h:p $MYVIMRC<cr>` / `:vsplit +lcd\ %:p:h $MYVIMRC<cr>` for tabs/windows respectively).

When switching between a project-window and config window, the harpoon list shows the relevant marks, i.e. project-related or config related marks, and the list stays persisted _after_ I quit and reopen neovim (I believe that was the original issue).

I will drop my own branch, and point to this in my lazy config until it gets merged.

lua/harpoon/init.lua
```diff
@@ -152,22 +157,28 @@ function Harpoon.setup(self, partial_config)
...
-                    if ev.event == "VimLeavePre" then
-                        self:sync()
+        vim.api.nvim_create_autocmd(
+            { "BufLeave", "VimLeavePre", "DirChanged" },
+            {
+                group = HarpoonGroup,
+                pattern = "*",
+                callback = function(ev)
+                    if ev.event == "DirChanged" then
+                        self:reset()
```
stroiman, line 168: Suggestion for minor technical improvement:

As far as I can tell from reading the code, this change seems to _reload the entire state_. The data model supports having lists multiple lists by working dir, so perhaps rather than completely reloading state, just load the state for the new dir _if it doesn't already exist in memory_, to optimize loading? Then the new file is only loaded _the first time_ you switch to a new directory.

Having said that, the harpoon file is so small that reading it in when changing tab or window is not a real problem. I don't experience any delays when using this branch, i.e. changing tabs is instantaneous.

cosmicboots: I'm inclined to leave it reloading from the disk completely. Reading from a file is fast enough the user shouldn't ever notice it, so it seems like a fairly negligible performance cost when considering what it would take to properly store all directory states in memory.

One big issue I could see with caching directory states in memory is garbage collection. When do we remove a directory's state from memory? Keeping them indefinitely is asking for memory leaks (albeit small) for anyone who a long running neovim instance when working with all their projects. It's also not obvious to me what a good heuristic would be to clear them out from memory (only keep 10..?, keep for a time..?).

Ultimately, I don't see the potential performance gains being worth complicating the state management logic further.

LehMichael:
This doesn't really work as expected (at least under the hood). After switching working directory, there are two lists stored, both with the default name. To make it work, you also need to set self.lists = {} in the reset method.
Other than that, it works great and i implemented the feature based on your code in my extension [harpoonEx](https://github.com/mike-jl/harpoonEx).

cosmicboots:
@mike-jl good catch! Fixed in 09e58f92339b7af3fd4f1d9c18a29448e8504282.

Tested and works from a user perspective as before, but now we're not dragging around old lists with us when we change directories.

carnage-mode:
How do we get this merged 😭

cosmicboots:
cc: @ThePrimeagen

PR #629, opened 2024-08-29
Fix nav_file column being off by 1

w1ck3dg0ph3r:
**What issue does it try to solve**
In `create_mark` and `store_offset` `vim.api.nvim_win_get_cursor` returns (1,0)-based cursor position, but `:call cursor()` in `nav_file` expects (1,1)-based one.


ThePrimeagen:
@w1ck3dg0ph3r ? More like w1ck3d smart

PR #631, opened 2024-09-14
fix(readme): sharing an easier way to install in lazyvim

vdegenne:


TheJoeSchr:
Not everyone that uses lazy.nvim package manager uses LazyVim as config setup. So either this changes should be added to the existing snippets or not added at all

vdegenne:
@TheJoeSchr Ah yes right, good catch. I'll change that

PR #644, opened 2024-11-12
added option to specify height in lines of the window

AmeerArsala:
Exactly what it sounds like. You can now specify how many lines the harpoon window should have.

What is my use-case for this? 

I want to be able to use more than 8 lines and want to see them all at once. I like to use 10, so I gave a general option for it.


ThePrimeagen:
Fine

AmeerArsala:
thanks brother 🔥🔥💯💯💯💯

Issue #653, opened 2025-01-24
[Harpoon2] list:determine_length seems broken when previous_length is 1


n8tlarsen:
I've been using Harpoon2 for a while now and added code in my personal config for terminal management (see #487). I've noticed that sometimes the an old terminal is selected, even though it has already been deleted. I did some digging and found that while the list:remove function is called (verified in the logs), the length of the list was still set to 1. So I took a look at the determine length function and while I don't know lua very well I think the function can never return zero unless the previous length was zero.
``` lua
local function determine_length(arr, previous_length)
    local idx = previous_length
    for i = previous_length, 1, -1 do
        if arr[i] ~= nil then
            idx = i
            break
        end
    end
    return idx
end
```

ThePrimeagen:
Fixed

PR #654, opened 2025-01-24
Fix for #653

n8tlarsen:


ThePrimeagen:
looking at this code
2 things.

Yes i ithnk you are right
2 shouldn't we just set idx to 0 instead of previous_length?

also we don't have a test.  i'll bring in this code and see if i can make this work!

PR #668, opened 2025-04-02
Fix: partial_config.settings.key() not working as expected

searleser97:
# Context
Currently if you add a custom "key()" function from the config, it is not being picked up by harpoon, because the initialization of the "data" variable is done before the "setup" call.

# What does this PR do?
it re-initializes the data variable during the setup call, so that the custom key function is picked up by harpoon


searleser97:
cc @ThePrimeagen

0xHolyTech:
Tested this change on my own system and solved the issue for me

searleser97:
cc @ThePrimeagen

PR #691, opened 2025-09-30
Feature/Fix: Handle swap files when opening files from Harpoon list

### LINKED ISSUE #570: [harpoon2] Native swap file error prompt suppressed by harpoon

### LINKED ISSUE #692: Harpoon2 does not handle Neovim's `E325 swap file found`.
### Description:
When opening a file from the Harpoon list, if Neovim detects a swap file, Harpoon2 does not handle it. Instead, Neovim shows the default swap file warning in the command line, which interrupts the otherwise smooth Harpoon workflow.

### Steps to Reproduce:
- **Open** a file in Neovim
- **Mark** it with Harpoon so it's in the Harpoon List
- **Close** the terminal with the file still open in Neovim (make sure you're not in a Tmux session)
- **Fire up** the terminal, open Nvim and select the file from the Harpoon list.
- **Withness** Neovim showing an unreadable `E325 swap file found` message and an error caused by the Harpoon plugin.

Image

### Expected Behavior:
Harpoon should present a consistent workflow when swap files are detected, rather than leaving the user with the raw Neovim warning.

### Actual Behavior:
The default Neovim swap file warning is shown, and Harpoon does not provide any handling for it.

### Simular Issues:
#570

Mathijs-Bakker:
### Problem:
Opening a file from the Harpoon list triggers Neovim’s swap file warning, but Harpoon2 does not handle it, which can be disruptive.

### Solution:
- **Intercept** swap files before opening the buffer using a wrapper around select().
- **Display** a Harpoon-style floating window with swap file information and user options: `(O)pen Read-only`, `(E)dit anyway`, `(R)ecover`, `(D)elete swap file`, `(A)bort`, `(Q)uit`.
- **Delegate** `(E)dit anyway` to Neovim to avoid data loss by this plugin (not our responsibility).

### Changes:
- `lua/harpoon/config.lua`
- Renamed `select()` to `_do_select()`
- Added wrapper `select()` to handle swap files
- `lua/harpoon/swap_ui.lua`
- Added floating window UI for swap file handling

### Tests:
- Existing tests behave the same as before. No test failures were introduced by these changes.

### Related Issue:
Closes #692 Harpoon2 does not handle Neovim's `E325 swap file found`.
Closes #570 [harpoon2] Native swap file error prompt suppressed by harpoon.

### Screenshots:

Before:
harpoon_e325_swap_issue

After:
harpoon_e325_fix

### Example Video:
[asciicast](https://asciinema.org/a/INvHfhT3pypICo9GIUbkDH8zV)

**Note:** the second time the `e` is selected (Edit anyway), this defaults to Neovim swap detection.
All other options are handled by Harpoon.


Mathijs-Bakker:
### Hold your horses. 

I've encountered an issue that should be resolved first:  
- **Selecting** (loaded) **buffers** from the list sometimes shows the **E325 swap_ui window** while selecting a buffers from the 
`Harpoon:list()`. Not always... but enough to be annoying.

Mathijs-Bakker:
**Fixed!**

ThePrimeagen:
this looks like ai generated non-sense..

PR #697, opened 2025-11-16
Add a Picker for Snacks.nvim

### LINKED ISSUE #696: Snacks.nvim picker for Harpoon (Code already done! Just don't know where it fits here)
I wrote a picker for [Snacks.nvim](https://github.com/folke/snacks.nvim) similar to the one for Telescope and thought it might be nice to add this to Harpoon directly.
It doesn't support editing the Harpoon entries like the Telescope picker since I never used this feature anyways.

I'm creating this issue since Snacks.nvim doesn't really have a similar extension mechanism like Telescope and I don't quite know where my picker would fit here.

Anyway, here's the code for the picker source:
```lua
local harpoon_source = {
	finder = function()
		local output = {}
		for _, item in ipairs(require("harpoon"):list().items) do
			if item and item.value:match("%S") then
				table.insert(output, {
					text = item.value,
					file = item.value,
					pos = { item.context.row, item.context.col },
				})
			end
		end
		return output
	end,
	filter = {
		transform = function()
			return true
		end,
	},
	format = function(item)
		return {
			{ item.text },
			{ ":", "SnacksPickerDelim" },
			{ tostring(item.pos[1]), "SnacksPickerRow" },
			{ ":", "SnacksPickerDelim" },
			{ tostring(item.pos[2]), "SnacksPickerCol" },
		}
	end,
	preview = function(ctx)
		if Snacks.picker.util.path(ctx.item) then
			return Snacks.picker.preview.file(ctx)
		else
			return Snacks.picker.preview.none(ctx)
		end
	end,
	confirm = "jump",
}
```

To use it with Snacks.nvim, simply set the following options and keybinds:
```lua
opts = {
	picker = {
		enabled = true,
		sources = { harpoon = harpoon_source },
	}
},
keys = {
		{
			"<Leader>H",
			function()
				Snacks.picker.pick("harpoon")
			end,
			desc = "Harpoon Marks",
		},
}
```

weisbrja:
This adds a picker for [Snacks.nvim](https://github.com/folke/snacks.nvim) which closes #696.
To enable the picker, install Snacks.nvim and run `require("harpoon.snacks").setup()`.

### DIFF
--- a/lua/harpoon/snacks.lua
+++ b/lua/harpoon/snacks.lua
@@ -1,0 +1,67 @@
+local M = {}
+
+local harpoon = require("harpoon")
+
+local changed = true
+local function notify()
+    changed = true
+end
+
+function M.setup()
+    if not Snacks then
+        error("harpoon requires folke/snacks.nvim")
+    end
+    Snacks.picker.sources.harpoon = {
+        finder = function()
+            local output = {}
+            for _, item in ipairs(harpoon:list().items) do
+                if item and item.value:match("%S") then
+                    table.insert(output, {
+                        text = item.value,
+                        file = item.value,
+                        pos = { item.context.row, item.context.col },
+                    })
+                end
+            end
+            return output
+        end,
+        filter = {
+            -- reruns the finder on changes
+            transform = function()
+                if changed then
+                    changed = false
+                    return true
+                end
+            end,
+        },
+        format = function(item)
+            return {
+                { item.text },
+                { ":", "SnacksPickerDelim" },
+                { tostring(item.pos[1]), "SnacksPickerRow" },
+                { ":", "SnacksPickerDelim" },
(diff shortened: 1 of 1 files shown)


xunafay:
Hey! Thanks for the PR it helped me kickstart my own setup for integrating snacks.nvim into harpoon. I did however come across some issues/missing features:
- Added a way to remove items from within the snacks picker
- Auto close the picker when it becomes empty
- Harpoon list doesn't always start at index 1, snacks.picker can't handle this so I had to modify the finder

My implementation: [https://github.com/xunafay/kickstart.nvim/blob/master/lua/custom/harpoon/snacks.lua](https://github.com/xunafay/kickstart.nvim/blob/master/lua/custom/harpoon/snacks.lua)
It's not the prettiest code as I am quite the lua/nvim noob, but it does work :P

image

PR #706, opened 2026-01-19
chore(workflows): fix the packages download path

ochcaroline:
## Summary
Fix the CI workflow so that it doesn't fail downloading packages

## Changes
* switch to new zip format available in GH packages (https://github.com/neovim/neovim/releases/tag/v0.11.5)
* move to newest nvim version for versioned download

## Reason
Since I had the same problem with my plugin, because I shamelessly copied your workflows - here are the results before and after.

Before:
image
image

After:
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,38 +3,42 @@ name: Tests
 on: [push, pull_request]
 
 jobs:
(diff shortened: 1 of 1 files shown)


zaier84 (approved):
looks good

